PR #3, opened 2021-07-19
CVL-14: Corrected RDS secret name

timharrison-moj:
Corrected database_user to database_username within the rds-instance-output secret.


yasinmustafa (approved):
LGTM

PR #16, opened 2021-09-21
CVSL-100: Added endpoint for updating initial appointment person

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
```diff
@@ -96,4 +99,47 @@ class LicenceController(private val licenceService: LicenceService) {
...
+      ),
+      ApiResponse(
+        responseCode = "404",
+        description = "The licence for this ID was not found.",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      )
+    ]
+  )
+  fun updateAppointmentPerson(
+    @PathVariable("licenceId") licenceId: Long,
```
timharrison-moj, line 140: Should there be a Void return type?

timharrison-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
smcveigh941: https://stackoverflow.com/questions/44834965/what-does-void-return-type-mean-in-kotlin

PR #17, opened 2021-09-23
CVSL-103: Added update endpoint for date/time of initial meeting.

timharrison-moj:
* Implements the update of the initial meeting date and time endpoint + tests + integration tests
* Combines the two `appointmentDate` and `appointmentTime` fields into one LocalDateTime in the model objects
* Combines the same two fields into a timestamp in the database table & entity
* Prior to committing this PR we need to remove the database tables & let them recreate via migration scripts in DEV.


src/main/resources/migration/common/V11__create_licence_tables.sql
```diff
@@ -41,8 +41,7 @@ CREATE TABLE licence (
   probation_area_code varchar(10),
   probation_ldu_code varchar(20),
   appointment_person varchar(60),
-  appointment_date date,
-  appointment_time varchar(8),
+  appointment_time timestamp with time zone,
```
mikehalmamoj, line 44: Is Flyway going to complain at this?

mikehalmamoj: Sorry just saw your comment on the PR - going to replace the DEV database. A bit of history rewriting there :smile:

mikehalmamoj (approved):

src/main/resources/migration/common/V11__create_licence_tables.sql
timharrison-moj: Keeps thing tidy - we're only in DEV at the moment. This luxury will go away soon. Thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AppointmentTimeRequest.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDateTime
+import javax.validation.constraints.NotNull
+
+@Schema(description = "Request object for updating the date and time of the initial appointment")
+data class AppointmentTimeRequest(
+  @Schema(description = "The date and time of the initial appointment", example = "12/12/2021 10:35")
+  @NotNull
```
smcveigh941, line 11: I think this needs to be `@field:NotNull`. Also, does this throw a `400: Bad request` error from the API if the date is in the wrong format/empty?

https://kotlinlang.org/docs/annotations.html#annotation-use-site-targets

smcveigh941 (changes requested):
Looks good, just have a question about how the validation works, will approve afterwards - thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/config/ControllerAdvice.kt
```diff
@@ -89,6 +90,20 @@ class ControllerAdvice {
...
+      .body(
+        ErrorResponse(
+          status = HttpStatus.BAD_REQUEST.value(),
+          userMessage = "Bad request: ${e.message}",
+          developerMessage = e.message
+        )
+      )
+  }
+
   @ExceptionHandler(ValidationException::class, MethodArgumentNotValidException::class)
```
smcveigh941, line 107: You could reuse the same controller advice by adding the exception type to this list, not sure if you want to do that

smcveigh941 (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/config/ControllerAdvice.kt
timharrison-moj: Consolidate on next PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AppointmentTimeRequest.kt
timharrison-moj: @field:NotNull - cant be used on a LocalDateTime unfortunately. I've added controller advice to catch JSON parse exceptions in requests to throw a 400 error + tests to verify.

PR #20, opened 2021-09-28
CVSL-101: Add endpoint to update appointment address

smcveigh941:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceControllerTest.kt
```diff
@@ -228,6 +229,31 @@ class LicenceControllerTest {
...
+
+  @Test
+  fun `update appointment address - invalid request body`() {
+    mvc.perform(
+      put("/licence/id/4/appointment-address")
+        .accept(APPLICATION_JSON)
+        .contentType(APPLICATION_JSON)
+        .content(mapper.writeValueAsBytes({ }))
+    )
+      .andExpect(status().isBadRequest)
```
mikehalmamoj, line 253: Possibly worth also having a test for a 404?

mikehalmamoj: Hmmm maybe not, I guess you're covered by the service test below (assuming there are other tests proving that `EntityNotFoundException` maps to a 404).

mikehalmamoj (approved):

PR #21, opened 2021-09-30
CVSL-111: Endpoint to update bespoke conditions for a licence

timharrison-moj:
New or empty bespoke conditions will entirely replace any bespoke conditions which currently exist on this licence.
This should satisfy the requirements to both create and update bespoke conditions for a licence.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
@@ -5,10 +5,12 @@ import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.AppointmentAddressRequest
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.AppointmentPersonRequest
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.AppointmentTimeRequest
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.BespokeConditionRequest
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -74,6 +78,21 @@ class LicenceService(
...
+  @Transactional
+  fun updateBespokeConditions(licenceId: Long, request: BespokeConditionRequest) {
+    licenceRepository
+      .findById(licenceId)
+      .orElseThrow { EntityNotFoundException("$licenceId") }
+
+    bespokeConditionRepository.deleteByLicenceId(licenceId)
+
+    request.conditions.forEachIndexed { index, condition ->
+      bespokeConditionRepository.saveAndFlush(
```
smcveigh941, line 90: Instead of saving and flushing all entities sequentially, you could build up a List of entities, and call `saveAllAndFlush(conditionsList)`

smcveigh941 (approved):
Looks good in general, suggestion about using `saveAll` using a list rather than looping through and saving sequentially

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
timharrison-moj: Yes, good call. Would you mind if I do that when I'm next in the API? I'm keen to get this in and crack on with the next bits in the UI, which needs the type imports from here.

smcveigh941: Sure thing

PR #29, opened 2021-10-19
CVSL-106: Update the list of additional conditions

smcveigh941:
- Refactored the ManyToOne mapping for additional conditions to make use of Hibernates ability to manage the referential integrity for us. Should probably do this for bespoke and standard conditions too...
- Left the additional data endpoint for now.. needs some more thinking about

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalCondition.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalCondition.kt
@@ -2,13 +2,14 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity
 
 import org.hibernate.annotations.Fetch
 import org.hibernate.annotations.FetchMode
-import org.hibernate.annotations.LazyCollection
-import org.hibernate.annotations.LazyCollectionOption
+import javax.persistence.CascadeType
 import javax.persistence.Entity
+import javax.persistence.FetchType
 import javax.persistence.GeneratedValue
 import javax.persistence.GenerationType.IDENTITY
 import javax.persistence.Id
 import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
 import javax.persistence.OneToMany
 import javax.persistence.OrderBy
 import javax.persistence.Table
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalCondition.kt
```diff
@@ -20,19 +21,18 @@ data class AdditionalCondition(
   @Id
   @GeneratedValue(strategy = IDENTITY)
   @NotNull
-  val id: Long = -1,
+  val id: Long? = null,
```
mikehalmamoj, line 24: I've just copied the `-1` solution because it means that the property is not nullable which reflects the real model... What was the reason for the switch to nullable?

smcveigh941: Reverted to the non-nullable field, thanks!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalConditionData.kt
```diff
@@ -15,8 +17,9 @@ data class AdditionalConditionData(
   @NotNull
   val id: Long = -1,
 
-  @NotNull
-  val additionalConditionId: Long = -1,
+  @ManyToOne
+  @JoinColumn(name = "additional_condition_id", nullable = false)
+  val additionalCondition: AdditionalCondition? = null,
```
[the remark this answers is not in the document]
smcveigh941: Sorted this, thanks

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/LicenceIntegrationTest.kt
```diff
@@ -451,6 +482,12 @@ class LicenceIntegrationTest : IntegrationTestBase() {
       conditions = listOf("Condition 1", "Condition 2", "Condition 3")
     )
 
+    val anAdditionalConditionsRequest = AdditionalConditionsRequest(
+      additionalConditions = listOf(
+        AdditionalCondition(code = "code", sequence = 0, text = "text")
+      )
```
timharrison-moj, line 488: Might be worth trying this with several additional conditions to ensure iterating over a number of them?

PR #36, opened 2021-11-08
CVSL-206: Stamp the full name of the approver on the licence.

timharrison-moj:
PLEASE DO NOT MERGE - requires table drop in DEV first.

This PR stamps the name of the approver into the licence table on APPROVAL so that it can be used for printing in the licence document as part of the signature page.


smcveigh941 (approved):
LGTM

PR #37, opened 2021-11-08
CVSL-184: Lengthen data field for additional conditions

smcveigh941:
- No restriction to the allowed length of the data field
- Requires database to be dropped to take effect


ldlharper (approved):
LGTM

PR #38, opened 2021-11-15
CVSL-227: Split standard and additional conditions to AP and PSS standard conditions

smcveigh941:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceControllerTest.kt
```diff
@@ -483,8 +483,10 @@ class LicenceControllerTest {
       probationLduCode = "LDU1",
       dateCreated = LocalDateTime.now(),
       createdByUsername = "X12345",
-      standardConditions = someStandardConditions,
-      additionalConditions = someAdditionalConditions,
+      standardLicenceConditions = someStandardConditions,
+      standardPssConditions = someStandardConditions,
+      additionalLicenceConditions = someAdditionalConditions,
+      additionalPssConditions = someAdditionalConditions,
```
timharrison-moj, line 489: Will this result in the same type of standard and additional conditions on a licence? ie. not separated into different types?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceControllerTest.kt
```diff
@@ -516,7 +518,8 @@ class LicenceControllerTest {
       comTelephone = "0116 2788777",
       probationAreaCode = "N01",
       probationLduCode = "LDU1",
-      standardConditions = someStandardConditions,
+      standardLicenceConditions = someStandardConditions,
+      standardPssConditions = someStandardConditions
     )
```
timharrison-moj, line 523: As above with this one too?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionsRequest.kt
```diff
@@ -8,4 +8,8 @@ data class AdditionalConditionsRequest(
   @Schema(description = "The list of additional conditions")
   @field:NotNull
   val additionalConditions: List<AdditionalCondition>,
+
+  @Schema(description = "The type of additional condition, either licence or post sentence supervision", allowableValues = ["AP", "PSS"])
+  @field:NotNull
```
timharrison-moj, line 13: Are we expecting there to be different standard and additional conditions by licence type? I don't know the answer to that, but just a thought. We might end up with convoluted config if there is no re-use?

timharrison-moj (approved):
LGTM. A nice and neat way to separate out the different condition types without a lot of complexity. Couple of random thoughts added.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionsRequest.kt
smcveigh941: My understanding is that there is one list of AP standard/additional conditions, and a list of PSS standard/additional conditions.

AP and PSS licences would have the respective condition types, and AP+PSS would have both

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceControllerTest.kt
smcveigh941: The licence object and the createLicence request object are agnostic of the `conditionType` field. It only splits them out into separate condition collections as they are here

PR #41, opened 2021-11-30
CVSL-174: Add endpoint to activate licences in bulk

smcveigh941:
New endpoint `/licence/activate-licences` which accepts a list of licence IDs, and changes the status of each of those licences to ACTIVE

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
@@ -21,6 +21,7 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.repository.LicenceR
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.repository.StandardConditionRepository
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.repository.toSpecification
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -231,6 +232,12 @@ class LicenceService(
     return transformToListOfSummaries(matchingLicences)
   }
 
+  fun activateLicences(licenceIds: List<Long>) {
+    val matchingLicences = licenceRepository.findAllById(licenceIds)
+    val activatedLicences = matchingLicences.map { licence -> licence.copy(statusCode = ACTIVE) }
+    licenceRepository.saveAllAndFlush(activatedLicences)
+  }
```
timharrison-moj, line 239: I wonder if we should do a check that these are still in an approved state?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/LicenceIntegrationTest.kt
```diff
@@ -481,6 +481,39 @@ class LicenceIntegrationTest : IntegrationTestBase() {
...
+
+    assertThat(result?.size).isEqualTo(3)
+    assertThat(result)
+      .extracting<Tuple> {
+        tuple(it.licenceId, it.licenceStatus)
+      }
+      .contains(
+        tuple(1L, LicenceStatus.ACTIVE),
+        tuple(2L, LicenceStatus.ACTIVE),
+      )
```
timharrison-moj, line 514: Expected 3 rows but checking 2?

timharrison-moj (approved):
Couple of comments, but minor only.

PR #42, opened 2021-11-30
CVSL-174: Skip activation of licences not in approved state

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -234,6 +234,11 @@ class LicenceService(
 
   fun activateLicences(licenceIds: List<Long>) {
     val matchingLicences = licenceRepository.findAllById(licenceIds)
+    matchingLicences.forEach { licence ->
+      if (licence.statusCode != APPROVED) {
+        throw ValidationException("Licence with ID ${licence.id} cannot be activated because it is not currently in the APPROVED state")
+      }
+    }
```
timharrison-moj, line 241: Fail the whole batch when 1 is not APPROVED? I know was my suggestion to check state, but it might be a bit drastic to fail all if one is not APPROVED?

timharrison-moj (commented):
Small comment about failing all because one is not APPROVED.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
smcveigh941: What would you suggest? Just log it and carry on?

timharrison-moj: Could we skip any that are not now approved? Or just ignore this, as its unlikely to happen between selecting the list and applying the activate.

smcveigh941: Pushed change, now skip any that are not approved but not fail

timharrison-moj (approved):
LGTM

PR #47, opened 2021-12-17
CVSL-274: Add submit licence endpoint which saves the COMs details

smcveigh941:
- Also removed the `getLicencesByStaffIdAndStatuses` endpoint as it's no longer used

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/SubmitLicenceRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/SubmitLicenceRequest.kt
@@ -1,0 +1,27 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+import io.swagger.v3.oas.annotations.media.Schema
+import javax.validation.constraints.NotBlank
+import javax.validation.constraints.NotNull
+
+@Schema(description = "Request object for submitting a licence")
+data class SubmitLicenceRequest(
+  @Schema(description = "The username of the person who is updating this status", example = "ssmyth")
+  @field:NotBlank
+  val username: String,
+
+  @Schema(description = "The DELIUS staff identifier of the person who is submitting status", example = "3000")
+  @field:NotNull
+  val staffIdentifier: Long,
+
+  @Schema(description = "The first name of the person who is submitting the licence", example = "John")
+  @field:NotBlank
+  val firstName: String,
+
+  @Schema(description = "The last name of the person who is submitting the licence", example = "Smythe")
+  @field:NotBlank
+  val surname: String,
+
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/SubmitLicenceRequest.kt
```diff
@@ -0,0 +1,27 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
```
timharrison-moj, line 2: Is this named clearly?  SubmitLicenceRequest that contains only the COMs details?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
```diff
@@ -9,6 +9,4 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
 @Repository
 interface LicenceRepository : JpaRepository<Licence, Long>, JpaSpecificationExecutor<Licence> {
   fun findAllByNomsIdAndStatusCodeIn(nomsId: String, status: List<LicenceStatus>): List<Licence>
-  fun findAllByComStaffIdAndStatusCodeIn(staffId: Long, status: List<LicenceStatus>): List<Licence>
```
timharrison-moj, line 12: Have you changed the way we find licences for a COM to now match on nomisId from the Delius caseload?

timharrison-moj: Its a good thing, by the way!

timharrison-moj (commented):
LGTM. Couple of questions only.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/SubmitLicenceRequest.kt
timharrison-moj: I think this is just my understanding - I can see that the com's details are not now added on creation of the licence, but on submission of the licence to prison - and this request therefore makes more sense.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
smcveigh941: Yes changed how it's queried, because the old way wouldn't account for an offenders COM changing

timharrison-moj (approved):
LGTM. We can see how it goes with the COM details added later in the process.

PR #50, opened 2021-12-21
CVSL-255 exclusion zone file upload, processing and retrieval of image and description.

timharrison-moj:
This PR:
* Allows the API to accept a multipart/form-data upload of the MapMaker PDF file.
* Extracts the map image from page 1 and the description text from page 2 
* Saves the PDF data and full size map image JPEG data into the `additional_condition_upload_detail` table 
* Saves the thumbnail JPEG image, descriptive text and file details into the `additional_condition_upload_summary` table.
* Links the summary table with additional conditions so it is available for each licence get
* Adds an endpoint to remove an upload - which removes both summary and detail rows, and unlinks from conditions.
* Adds an endpoint to get the full-size map image as content-type image/jpeg for printing in the licence appendices.
* Unit and integration tests for the above features.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalConditionUploadDetail.kt
```diff
@@ -0,0 +1,53 @@
...
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  @NotNull
+  val id: Long = -1,
+
+  @NotNull
+  val licenceId: Long,
+
+  @NotNull
+  val additionalConditionId: Long,
```
smcveigh941, line 24: Don't think you need a relationship between the additional_condition table and the licence_table? The licence can already be referenced via the additional condition?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalConditionUploadDetail.kt
```diff
@@ -0,0 +1,53 @@
...
+
+  @Lob
+  @Column(name = "full_size_image", columnDefinition = "BLOB")
+  val fullSizeImage: ByteArray? = null,
+) {
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (javaClass != other?.javaClass) return false
+
+    other as AdditionalConditionUploadDetail
```
smcveigh941, line 38: Is this line needed as the above line will kick us out of here if `other` isnt of the correct type?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalConditionUploadSummary.kt
```diff
@@ -0,0 +1,75 @@
...
+@Table(name = "additional_condition_upload_summary")
+data class AdditionalConditionUploadSummary(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  @NotNull
+  val id: Long = -1,
+
+  @ManyToOne
+  @JoinColumn(name = "additional_condition_id", nullable = false)
+  val additionalCondition: AdditionalCondition,
```
smcveigh941, line 25: This one doesn't have the licenceId as foreign key - which I think is correct - but not consistent with the UploadDetail

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionUploadSummary.kt
```diff
@@ -0,0 +1,61 @@
...
+
+  @Schema(description = "The thumbnail image for the image map contained in this document", example = "ByteArray")
+  val thumbnailImage: ByteArray? = null,
+
+  @Schema(description = "The id which references the original file data and full size image", example = "9999")
+  val uploadDetailId: Long,
+) {
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (javaClass != other?.javaClass) return false
```
smcveigh941, line 36: Are these equals and hash functions ever used?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/ExclusionZoneController.kt
```diff
@@ -0,0 +1,156 @@
...
+  )
+  fun uploadExclusionZoneFile(
+    @PathVariable(value = "licenceId") licenceId: Long,
+    @PathVariable(value = "conditionId") conditionId: Long,
+    @RequestPart("file") file: MultipartFile
+  ) {
+    try {
+      exclusionZoneService.uploadExclusionZoneFile(licenceId, conditionId, file)
+    } catch (e: IOException) {
+      throw ValidationException("Exclusion zone file could not be processed")
```
smcveigh941, line 71: This can be handled in the ControllerAdvice by specifying a response for type `IOException`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
```diff
@@ -0,0 +1,182 @@
...
+    return ByteArray(0)
+  }
+
+  fun extractDescription(fileStream: InputStream): String? {
+    var pdfDoc: PDDocument? = null
+    try {
+      pdfDoc = PDDocument.load(fileStream)
+      val stripper = PDFTextStripper()
+      stripper.sortByPosition = true
+      stripper.startPage = 2
```
smcveigh941, line 143: Can you be sure that the description will be on page 2? Never on page 1 even?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
```diff
@@ -0,0 +1,182 @@
...
+      return stripper.getText(pdfDoc)
+    } catch (e: Exception) {
```
smcveigh941, line 145: Are you able to be more specific with which type of exception might throw here? Catching general exceptions is considered bad practice - https://newbedev.com/is-it-really-that-bad-to-catch-a-general-exception#:~:text=Unless%20you%20are,everything%20is%20ok.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalConditionUploadDetail.kt
timharrison-moj: This `additional_condition_upload_detail` table has no JPA relationships - it is entirely managed manually, from the earlier discussion about pulling potentially large objects for each licence if we link the detail row in. The licenceId column is purely there for our benefit (for support/visibility) so we can query which licences have an upload detail associated with them. It isn't used in getting data or managing any relationships between tables.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalConditionUploadDetail.kt
timharrison-moj: Its generated code. The only reason for the equals and hashcode functions is remove the IntelliJ warnings. When a class contains Blobs it suggests to override the hashcode, equals and toString functions to avoid the potentially long binary columns. They're just generated to ignore the the BLOB columns.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalConditionUploadSummary.kt
timharrison-moj: Yes - this one does have the relationship managed and is associated with an `additionalCondition` entity.  The uploadDetail is effectively stand-alone, and not managed via any JPA relationships.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionUploadSummary.kt
timharrison-moj: Not sure - they are generated functions to remove the warnings given by IntelliJ (as above).

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/ExclusionZoneController.kt
timharrison-moj: Yes - it might better be handled within the ExclusionZoneService rather than thrown up to here. I'll take a look at this - and its related to a comment below, about catching the generic exception, so can address both of these together.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
timharrison-moj: Yes!  I don't like it, as its very specific to this file created by MapMaker, which always puts the large area map on Page 1 and the descriptive text on Page 2.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
timharrison-moj: Yes - as mentioned above - I can take a look at these exceptions when processing PDFs and Images. They can be more specific.

smcveigh941: Looks better now, thanks

smcveigh941 (approved):

PR #54, opened 2022-01-11
CVSL-313: Revert to page render for map. Crop the borders away.

timharrison-moj:
Extracting the full-size image from the PDF produces a better/bigger image, but loses the zone overlay shape - which is kind of the point of the map! This change reverts to rendering page-1 of the PDF as an image, but then cropping away the borders so that we produce a reasonable map + overlay zone.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
@@ -2,7 +2,7 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
 
 import org.apache.pdfbox.pdmodel.PDDocument
 import org.apache.pdfbox.pdmodel.encryption.InvalidPasswordException
-import org.apache.pdfbox.pdmodel.graphics.image.PDImageXObject
+import org.apache.pdfbox.rendering.PDFRenderer
 import org.apache.pdfbox.text.PDFTextStripper
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
```diff
@@ -146,15 +146,15 @@ class ExclusionZoneService(
...
-          return baos.toByteArray()
-        }
-      }
+      val renderer = PDFRenderer(pdfDoc)
+      val firstImage = renderer.renderImage(0)
+      val croppedImage = firstImage.getSubimage(50, 50, firstImage.width - 100, firstImage.height - 100)
+      val baos = ByteArrayOutputStream()
+      ImageIO.write(croppedImage, "jpg", baos)
+
+      // Scale it back up to original size without the borders ?
```
smcveigh941, line 155: Is this a TODO?

smcveigh941 (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
timharrison-moj: Just a thought for now - didn't intend to leave it it. It looks ok as is for now, but I think there are a couple of other things we could do to make the map sharper. I've created a ticket and put into the backlog for these.  1) Alter the format from JPEG to PNG (lossless compression) - for maps with lines and small text, this might really help with clarity - depending on the quality of printers in prisons!. 2) Scale the image up slightly so it fills more of the page.  A combination of both would improve it - but I'd rather wait and see what the feedback is?

PR #56, opened 2022-01-14
CVSL-326: Added helm and circle-ci config for preprod.

timharrison-moj:


helm_deploy/values-preprod.yaml
```diff
@@ -0,0 +1,18 @@
+---
+# Per environment values which override defaults in create-and-vary-a-licence-api/values.yaml
+
+generic-service:
+  replicaCount: 2
```
smcveigh941, line 5: Same number of replicas as dev.. is this standard for preprod and prod?

timharrison-moj: Usually preprod mirrors prod, but I thought it might be a bit wasteful of resources? We don't actually use preprod for much since its hard to access - people don't generally have accounts to get on it. I checked some other services and there is a mix  - some 4, some 2, probably for similar reasons.

timharrison-moj: Its something we can alter quickly if we planned a performance test.

smcveigh941 (approved):

PR #57, opened 2022-01-20
CVSL-339: Import kotlin JPA specification directly

timharrison-moj:
This is no longer available from the JCenter repo now.
Imported the single file (Apache 2.0) licence conditions.


smcveigh941 (approved):
LGTM

PR #58, opened 2022-01-24
CVSL-36: Add govuk-notify into the service.

timharrison-moj:
* Adds govuk notify dependency, helm values and properties.
* Only a single template exists for licence approved - which emails the COM.
* Tests for notify service
* Removed all test-data features (controller, service, repository, tests, tables, data loads etc)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/TestData.kt
```diff
@@ -1,14 +0,0 @@
-package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity
```
smcveigh941, line 1: 😍

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/NotifyService.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Service
+import uk.gov.service.notify.NotificationClient
+
+@Service
+class NotifyService(
+  @Value("\${notify.enabled}") private val enabled: Boolean,
+  @Value("\${notify.templates.licence-approved-id}") private val licenceApprovedTemplateId: String,
```
smcveigh941, line 11: Not sure that passing these `@Value`s as a parameter is the right thing to. I think usually they would just be class properties. If the reason for passing them in as a parameter is to help with unit tests, you can use ReflectionTestUtils which mock the `application.yml` values

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/config/NotifyConfiguration.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import uk.gov.service.notify.NotificationClient
+
+@Configuration
+class NotifyConfiguration(@Value("\${notify.api.key:invalidKey}") private val apiKey: String) {
+  @Bean
+  fun notifyClient(): NotificationClient {
```
smcveigh941, line 11: How does the client communicate with the API? I don't see a URL anywhere

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/NotifyService.kt
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    if (emailAddress.isNullOrEmpty()) {
+      log.info("Blank email address: Did not send notification for $templateId ref $reference")
+      return
+    }
+
+    try {
+      client.sendEmail(templateId, emailAddress, values, reference)
+    } catch (e: Exception) {
```
smcveigh941, line 31: Do you know which Exceptions this can throw? Catching generic Exception can cover up real problems

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/config/NotifyConfiguration.kt
timharrison-moj: Think this is built into the java client as a default value. It can take a proxy, but I don't think we need one. e.g. `NotificationClient client = new NotificationClient(apiKey, proxy);`

src/main/resources/application.yml
```diff
@@ -89,3 +89,8 @@ management:
     info:
       cache:
         time-to-live: 2000ms
+
+notify:
+  enabled: true
+  templates:
+    licence-approved-id: 36c3f270-7393-4595-96f7-db2033cccd91
```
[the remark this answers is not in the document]
timharrison-moj: Yes - it will be different in each. These are just default values, for now. As above, proper place will the helm env-specific yaml files.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/NotifyService.kt
timharrison-moj: Yes, they are all child exceptions of `NotifyClientException` so will alter to be more specific.

src/main/resources/application.yml
```diff
@@ -89,3 +89,8 @@ management:
     info:
       cache:
         time-to-live: 2000ms
+
+notify:
+  enabled: true
+  templates:
+    licenceApproved: 36c3f270-7393-4595-96f7-db2033cccd91
```
smcveigh941, line 96:
```suggestion
    licenceApproved: ${NOTIFY_TEMPLATES_LICENCEAPPROVED}
```

timharrison-moj: I left a default value in the application.yaml (for template IDs) is so that it will default to this when running locally, without having to set an environment variable and would be overridden by values from Helm in real environments - dev, preprod and prod.

smcveigh941 (approved):
LGTM

PR #59, opened 2022-01-25
CVSL-341: Add COM Username to LicenceSummary

smcveigh941:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -705,14 +705,14 @@ class LicenceServiceTest {
       topupSupervisionExpiryDate = LocalDate.of(2021, 10, 22),
       comFirstName = "Stephen",
       comLastName = "Mills",
```
timharrison-moj, line 707: Should we remove the first/last names everywhere, and maybe drop the columns from the licence table?  Are they used anywhere now?

timharrison-moj (approved):
LGTM. One question only.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
smcveigh941: They're not used. I think there is a task to clean up the DB schema in general

smcveigh941: Will put one in as `tech debt`

PR #60, opened 2022-01-28
CVSL-295: Endpoint to update the COM allocated to an offender who has a licence in flight

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/UpdateResponsibleComRequest.kt
```diff
@@ -0,0 +1,20 @@
...
+  @field:NotNull
+  val staffIdentifier: Long,
+
+  @Schema(description = "The Delius username for the responsible COM", example = "jbloggs")
+  @field:NotNull
+  val staffUsername: String,
+
+  @Schema(description = "The email address of the responsible COM", example = "jbloggs@probation.gov.uk")
+  @field:NotNull
+  val staffEmail: String,
```
timharrison-moj, line 19: Is it worth adding a telephone number here? Or leave until we see the need?

smcveigh941: Was just adding the details needed right now. Email is needed for notify to work, but we will have reconsiderations when we create the COM table I'm sure

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/OffenderService.kt
```diff
@@ -0,0 +1,24 @@
...
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.UpdateResponsibleComRequest
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.repository.LicenceRepository
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
+
+@Service
+class OffenderService(
+  private val licenceRepository: LicenceRepository,
+) {
+
+  fun updateOffenderWithResponsibleCom(crn: String, newComDetails: UpdateResponsibleComRequest) {
```
timharrison-moj, line 13: Should probably be @Transactional?  Maybe add a TODO here to record an audit event associated with this licence?

timharrison-moj (commented):
LGTM. Minor comment only.

timharrison-moj (approved):
LGTM

PR #61, opened 2022-02-01
CVSL-361 add auditing features

timharrison-moj:
*  Audited events can be related to a user, a licence, or both.
* They may not be in relation to a change to data - e.g. could be for when a user views a licence.
* They may not be related to a licence - so no referential integrity to licence data.
* Endpoints to create an audit event, get events for a user, get events for a licence, and get all events
* Events are queried between two dates (defaulted to 1 month if not supplied).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonFormat
 import io.swagger.v3.oas.annotations.media.Schema
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.AuditEventType
 import java.time.LocalDateTime
+import javax.validation.constraints.NotEmpty
 
 @Schema(description = "Describes an audit event request")
 data class AuditEvent(
@@ -18,15 +19,16 @@ data class AuditEvent(
   val eventTime: LocalDateTime = LocalDateTime.now(),
 
   @Schema(description = "Username who initiated the event, if a user event, or SYSTEM if an automated event", example = "X63533")
-  val username: String? = null,
+  val username: String? = "SYSTEM",
 
   @Schema(description = "The full name of the person who performed this auditable event, or SYSTEM if an automated event.", example = "Robert Mortimer")
-  val fullName: String? = null,
+  val fullName: String? = "SYSTEM",
 
   @Schema(description = "The event type. One of SYSTEM_EVENT or USER_EVENT", example = "USER_EVENT")
   val eventType: AuditEventType = AuditEventType.USER_EVENT,
 
   @Schema(description = "A summary of the action taken", example = "Updated a bespoke condition")
+  @NotEmpty
   val summary: String? = null,
 
   @Schema(description = "A detailed description of the action taken", example = "Updated a bespoke condition")
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
@@ -1,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.AuditEventType
+import java.time.LocalDateTime
+
+@Schema(description = "Describes an audit event request")
+data class AuditEvent(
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  @Schema(description = "The internal ID of the licence that this event related to, or null if unrelated to a licence", example = "1234")
+  val licenceId: Long? = null,
+
+  @Schema(description = "The date and time of the event", example = "12/01/2022 23:14:23")
+  @JsonFormat(pattern = "dd/MM/yyyy HH:mm:ss")
+  val eventTime: LocalDateTime = LocalDateTime.now(),
+
+  @Schema(description = "Username who initiated the event, if a user event, or SYSTEM if an automated event", example = "X63533")
+  val username: String? = null,
```
smcveigh941, line 21: Could it default to SYSTEM if username is not supplied?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  @Schema(description = "The full name of the person who performed this auditable event, or SYSTEM if an automated event.", example = "Robert Mortimer")
+  val fullName: String? = null,
+
+  @Schema(description = "The event type. One of SYSTEM_EVENT or USER_EVENT", example = "USER_EVENT")
+  val eventType: AuditEventType = AuditEventType.USER_EVENT,
+
+  @Schema(description = "A summary of the action taken", example = "Updated a bespoke condition")
+  val summary: String? = null,
```
smcveigh941, line 30: No validation on this field? A summery of the event would always be required right? Might need `@field::NotBlank`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.AuditEventType
+import java.time.LocalDateTime
+
+@Schema(description = "Describes an audit event request")
+data class AuditEvent(
+  @Schema(description = "The internal ID of the audit event", example = "1234")
+  val id: Long? = null,
```
smcveigh941, line 11: I don't generally like using the same DTO for request and response. The request to create the audit would not have an ID supplied, and may actually cause error if the ID _was_ supplied. Whereas the response will have an ID returned. The request does not look the same as the response and I would usually split the DTOs into request DTO and response DTO.

Although, this is a wider issue with the API, not just related to this

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
```diff
@@ -0,0 +1,69 @@
...
+class AuditService(
+  private val auditEventRepository: AuditEventRepository,
+  private val licenceRepository: LicenceRepository,
+) {
+
+  fun recordAuditEvent(auditEvent: ModelAuditEvent) {
+    auditEventRepository.save(transform(auditEvent))
+  }
+
+  fun getAuditEvents(auditRequest: AuditRequest): List<ModelAuditEvent> {
```
smcveigh941, line 20: `@Transactional`?

smcveigh941 (dismissed):
Couple of comments but happy in general

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
timharrison-moj: Yes, I will add this.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
timharrison-moj: Yes, will add.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AuditEvent.kt
timharrison-moj: The same object is used as its identical (almost!). The ID is nullable for requests, so made sense for convenience, and keeping it simple. This might evolve - I'm already wondering about whether I should record specific event_types e.g. login, logout, create, approve, vary, edit, print, view etc.. Might be useful to be able to query all events for create etc.  Could also include the user type (prison or probation), and which prison or team ... but for now, I thought I'd do the minimum, and we can evolve the needs arise.

smcveigh941 (approved):

PR #62, opened 2022-02-04
CVSL-384: Update gradle plugin

timharrison-moj:
Fix vulnerability in Tomcat reported overnight.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.0.1"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.0.3-beta"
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.0.1"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.0.3-beta"
```
smcveigh941, line 2: Think this may break when the tested version has been released?

smcveigh941 (approved):

build.gradle.kts
timharrison-moj: I think the 4.0.3 will go out soon, and both versions will still be deployed and available AFAIK.

PR #65, opened 2022-02-09
CVSL-349: Add variation statuses

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
```diff
@@ -13,15 +13,14 @@ data class Licence(
...
+  @Schema(description = "The licence type code", example = "AP")
   val typeCode: LicenceType,
 
   @Schema(description = "The version number used for standard and additional conditions", example = "1.4")
   val version: String? = null,
 
   @Schema(
     description = "The current status code for this licence",
-    allowableValues = ["IN_PROGRESS", "SUBMITTED", "APPROVED", "REJECTED", "ACTIVE", "INACTIVE", "RECALLED"],
     example = "IN_PROGRESS",
```
timharrison-moj, line 24: Do we want to take these off? Or add more? Or temporarily remove and then add them all back in when they're stable?

PR #68, opened 2022-02-11
CVSL-402: Update to plugin 4.0.3 and tidy unnecessary null check

timharrison-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
```diff
@@ -57,11 +57,7 @@ fun hasNomsIdIn(nomsIds: List<String>?): Specification<Licence>? = nomsIds?.let
 
 fun hasResponsibleComIn(staffIds: List<Int>?): Specification<Licence>? = staffIds?.let {
   return Specification<Licence> { root, _, builder ->
-    if (staffIds.isNullOrEmpty()) {
```
smcveigh941, line 60: Ah, because it wouldn't go in here in the first place if there was nothing in the list

smcveigh941 (approved):

PR #83, opened 2022-02-25
CVSL-394:  Variation approval and rejection/referral endpoints

timharrison-moj:
* Adds an endpoint to approve a variation
* Adds an endpoint to refer a variation
* Adds PDU to the list of licence matching criteria
* Unit tests to cover the above
(Integration tests to follow)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
@@ -79,6 +79,7 @@ data class Licence(
   val spoDiscussion: String? = null,
   val vloDiscussion: String? = null,
   val reasonForVariation: String? = null,
+  val reasonForReferral: String? = null,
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
```diff
@@ -79,6 +79,7 @@ data class Licence(
   val spoDiscussion: String? = null,
   val vloDiscussion: String? = null,
   val reasonForVariation: String? = null,
+  val reasonForReferral: String? = null,
```
smcveigh941, line 82: So going with only one reason initially?

timharrison-moj: That's the plan. Then evolve so `licence_history` becomes `licence_event` and we can add different event types to reflect the `reasonForVariation` and `reasonForReferral` conversation between COM and ACO. Still room for discussion over on how that works.

PR #84, opened 2022-02-28
Fix deletion of bespoke conditions

smcveigh941:
Because the bespoke conditions are now joined by hibernate, instead of a manual join by id, the deletion by ID no longer works. Instead, the entity should be updated with an empty list

Infact, the BespokeConditionRepository should not exist, and should be handled through the licence repository as a join, but this is a bigger piece of work to refactor

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
@@ -188,11 +188,10 @@ class LicenceService(
       .orElseThrow { EntityNotFoundException("$licenceId") }
 
     val username = SecurityContextHolder.getContext().authentication.name
-    val updatedLicence = licenceEntity.copy(dateLastUpdated = LocalDateTime.now(), updatedByUsername = username)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -188,11 +188,10 @@ class LicenceService(
       .orElseThrow { EntityNotFoundException("$licenceId") }
 
     val username = SecurityContextHolder.getContext().authentication.name
-    val updatedLicence = licenceEntity.copy(dateLastUpdated = LocalDateTime.now(), updatedByUsername = username)
+    val updatedLicence = licenceEntity.copy(bespokeConditions = emptyList(), dateLastUpdated = LocalDateTime.now(), updatedByUsername = username)
     licenceRepository.saveAndFlush(updatedLicence)
 
     // Replace the bespoke conditions
-    bespokeConditionRepository.deleteByLicenceId(licenceId)
     request.conditions.forEachIndexed { index, condition ->
```
timharrison-moj, line 195: Should these be set into a list on the Licence entity too?

smcveigh941: Yes, the BespokeConditionRepository should not exist, as they should be handled by LicenceRepository using joins within Hibernate. But this works ok as is for now - until such times as a wider refactor is done

PR #85, opened 2022-03-01
CVSL-377 prod helm API values

timharrison-moj:
* CircleCI pipeline to push to production.
* `values-prod.yaml` configuration.

### DIFF
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -1,0 +1,19 @@
+---
+# Per environment values which override defaults in create-and-vary-a-licence-api/values.yaml
+
+generic-service:
+  replicaCount: 4
+
+  ingress:
+    host: create-and-vary-a-licence-api.hmpps.service.justice.gov.uk
+    contextColour: green
+
+  env:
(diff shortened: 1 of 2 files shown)


helm_deploy/values-prod.yaml
```diff
@@ -0,0 +1,19 @@
...
+  replicaCount: 4
+
+  ingress:
+    host: create-and-vary-a-licence-api.hmpps.service.justice.gov.uk
+    contextColour: green
+
+  env:
+    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    HMPPS_AUTH_URL: https://sign-in.hmpps.service.justice.gov.uk/auth
+    NOTIFY_TEMPLATES_LICENCEAPPROVED: 36c3f270-7393-4595-96f7-db2033cccd91
```
smcveigh941, line 14: Same template used in dev? Should they be separate so that they can be tested in dev before promoting to prod?

timharrison-moj: Its the best we have currently, and still configured only to send where the email address is within the team. Need to think about how best to separate the email templates for DEV/PREPROD and PROD.

smcveigh941 (approved):
LGTM

PR #89, opened 2022-03-08
CVSL-422:  Change licence history to events, record events and add endpoint to query them

timharrison-moj:
PLEASE DO NOT MERGE - there will be an accompanying UI change to go with this.

* Removes the licence_history table
* Adds a licence_event table
* Records events at all the appropriate places we might be interested in
* Records the variation reason and rejection reasons
* Endpoint to search for events for a licence / all licences / event types.
* Tests and integration tests.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/OffenderService.kt
```diff
@@ -3,22 +3,40 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
...
     val inFlightLicenceStatuses = listOf(LicenceStatus.IN_PROGRESS, LicenceStatus.SUBMITTED, LicenceStatus.APPROVED, LicenceStatus.ACTIVE)
-
     var offenderLicences = this.licenceRepository.findAllByCrnAndStatusCodeIn(crn, inFlightLicenceStatuses)
 
+    // Update the active licences for this person on probation
     offenderLicences = offenderLicences.map { it.copy(responsibleCom = newCom) }
-
-    // TODO: Create an audit log
-
     this.licenceRepository.saveAllAndFlush(offenderLicences)
```
timharrison-moj, line 25: Implemented this TODO

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/LicenceIntegrationTest.kt
```diff
@@ -585,16 +586,18 @@ class LicenceIntegrationTest : IntegrationTestBase() {
       .expectStatus().isOk
 
     val result = webTestClient.get()
-      .uri("/licence/id/1")
+      .uri("/events/match?licenceId=1&eventType=VARIATION_SUBMITTED_REASON")
       .accept(MediaType.APPLICATION_JSON)
```
timharrison-moj, line 590: Checks that the reason was record in licence_events, not now on the licence itself.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceEvent.kt
```diff
@@ -0,0 +1,46 @@
...
+    example = "Robert Mortimer"
+  )
+  val surname: String? = null,
+
+  @Schema(description = "A reason or description related to the event", example = "Reason for variation")
+  val eventDescription: String? = null,
+
+  @Schema(description = "The date and time of the event", example = "12/01/2022 23:14:23")
+  @JsonFormat(pattern = "dd/MM/yyyy HH:mm:ss")
+  val eventTime: LocalDateTime = LocalDateTime.now(),
```
smcveigh941, line 45: This should not have a default value. If there is a problem with the database where this value was not populated, we want to know about it by letting it fail, and we shouldnt cover it up by returning an inaccurate value

src/test/resources/test_data/clear-all-data.sql
```diff
@@ -4,7 +4,6 @@ DELETE from additional_condition_upload_detail;
 DELETE from additional_condition_upload_summary;
 DELETE from additional_condition;
 DELETE from bespoke_condition;
-DELETE from licence_history;
```
smcveigh941, line 7: Want to delete the licence events instead?

smcveigh941 (commented):
Looks like good work. I believe that eventTime shouldnt have a default value on the DTO though

src/test/resources/test_data/clear-all-data.sql
timharrison-moj: No, its ok, there's a cascading delete for events when deleting the licence.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceEvent.kt
timharrison-moj: Thanks - I'll have a check on this next.

src/test/resources/test_data/clear-all-data.sql
timharrison-moj: Audit events do not cascade deletes - we keep this regardless, but licence events are always linked with a licence and do.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceEvent.kt
timharrison-moj: Yes, you're right. The model object is only used to return licence events from the DB, so if there's no date and time specified there we should return no date and time to the client. Changing this now.

timharrison-moj: Updated now.

smcveigh941 (approved):
LGTM

PR #96, opened 2022-03-16
CVSL-444: Record digital signature of variation approver

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -683,6 +684,9 @@ class LicenceService(
       statusCode = ACTIVE,
       dateLastUpdated = LocalDateTime.now(),
       updatedByUsername = username,
+      approvedByUsername = username,
+      approvedDate = LocalDateTime.now(),
+      approvedByName = body.fullName
     )
```
timharrison-moj, line 690: Should the full name of the approver also go into the events below this? Looking at it, its assuming that the approver will also appear in the comDetail table - recorded at login, so it looks like the full name is already obtainable from there?

timharrison-moj (commented):
LGTM. One question about whether the additional body content of full name is necessary, or already obtainable from comDetail?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
smcveigh941: Sorted this now, thanks

timharrison-moj (approved):
LGTM

PR #99, opened 2022-03-23
CVSL-457 Variation approval status

timharrison-moj:
3 minor changes:
* When a licence variation is approved by the ACO the status is changed to VARIATION_APPROVED instead of ACTIVE.
* I've added an index to the licence column `variation_of_id` so that we can quickly find discarded variations later.
* Added a new field `createdByFullName` into the licence model object, and populated this from the createdBy join column.


smcveigh941 (approved):
LGTM, the race is on for #100

PR #120, opened 2022-04-30
CVSL-464: Choosing SYSTEM or USER event more correctly in auditing

timharrison-moj:
Small change - if no user present, or user.username == 'SYSTEM' then select the SYSTEM_EVENT for audit, otherwise USER_EVENT.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
@@ -355,11 +355,20 @@ class LicenceService(
           licenceId = licenceId,
           username = request.username,
           fullName = request.fullName,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -355,13 +355,22 @@ class LicenceService(
...
   }
 
+  private fun getAuditEventType(request: StatusUpdateRequest): AuditEventType {
+    return if (request.username == "SYSTEM") {
+      AuditEventType.SYSTEM_EVENT
+    } else {
+      AuditEventType.USER_EVENT
+    }
+  }
+
```
smcveigh941, line 373: This username _**can**_ be `SYSTEM` because thats what we pass into the endpoint

smcveigh941 (approved):
LGTM

PR #121, opened 2022-05-03
CVSL-541: Update probation team on licence when offender manager is updated

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
```diff
@@ -62,4 +64,45 @@ class OffenderController(private val offenderService: OffenderService, private v
...
+      )
+    ]
+  )
+  @Transactional
+  fun updateProbationTeam(
+    @PathVariable crn: String,
+    @RequestBody body: UpdateProbationTeamRequest
+  ) {
+    this.offenderService.updateProbationRegion(crn, body)
+  }
```
timharrison-moj, line 107: Just naming potentially? `updateProbationTeam` calling a service method called `updateProbationRegion`? Probation region is a specific value. Small one, not a blocker.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/OffenderService.kt
```diff
@@ -39,4 +40,47 @@ OffenderService(
       )
     }
   }
+
+  @Transactional
+  fun updateProbationRegion(crn: String, request: UpdateProbationTeamRequest) {
+    val inFlightLicenceStatuses = listOf(LicenceStatus.IN_PROGRESS, LicenceStatus.SUBMITTED, LicenceStatus.APPROVED, LicenceStatus.ACTIVE)
+    var offenderLicences = this.licenceRepository.findAllByCrnAndStatusCodeIn(crn, inFlightLicenceStatuses)
```
timharrison-moj, line 47: Should we include VARIATION_IN_PROGRESS, VARIATION_SUBMITTED and VARIATION_APPROVED / REJECTED too?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
smcveigh941: Sorted now, thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/OffenderService.kt
smcveigh941: Oh! Yes we should, looks like we missed it on updating the COM too. Fixed now

timharrison-moj (approved):
LGTM

PR #128, opened 2022-05-17
CVSL-546: Ensure the licence created_time is populated + gradle plugin update.

timharrison-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -83,6 +83,8 @@ class LicenceService(
       ?: throw RuntimeException("Staff with username $username not found")
 
     val licence = transform(request)
+
+    licence.dateCreated = LocalDateTime.now()
```
mikehalmamoj, line 87: Just wondering why not do this in the constructor? Then the property is immutable...

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
timharrison-moj: Its a good question - could be done in the transform(request) - but there are other things which the API enriches after creating the licence object so for simplicity I added to the same section to do them all together.

PR #129, opened 2022-05-19
CVSL-410: Notifications for licence variation approval or referral

timharrison-moj:
When approved or referred, a notification is sent to the interested parties.

These are:
  * The responsible COM
  * The person who submitted the variation (it could be somebody different)

If the two are the same, only one notification is generated.
If they are different, each get the notification.
If only one of them is present - they get the notification.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -727,6 +735,15 @@ class LicenceService(
         )
       )
     )
+
```
mgreen-mojhub, line 738: Is it possible that the createdBy AND COM Email are null? If so, it might be an idea to check if both are messing before calling the method, and logging the lack of email addresses using the logger before calling return on the current method.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/NotifyService.kt
```diff
@@ -113,18 +161,40 @@ class NotifyService(
...
+  private fun getContacts(creatorEmail: String, creatorName: String, comEmail: String, comName: String): List<Contact> {
+    val contacts = mutableListOf<Contact>()
+    val creatorIsBlank = creatorEmail.isEmpty()
+    val comIsBlank = comEmail.isEmpty()
+    if (creatorIsBlank && !comIsBlank) {
+      contacts.add(Contact(comEmail, comName))
+    } else if (comIsBlank && !creatorIsBlank) {
+      contacts.add(Contact(creatorEmail, creatorName))
+    } else if (creatorEmail == comEmail) {
+      contacts.add(Contact(creatorEmail, creatorName))
```
mgreen-mojhub, line 184: You could combine this into one statement:
else if((comIsBlank && !creatorIsBlank) || creatorEmail == comEmail)

It saves a duplicate line of code, but adds more complexity to the if statement. So quite subjective.

timharrison-moj: Yes, I thought of doing it like that, but settled on the above way and being 'easier to read'.

mgreen-mojhub (approved):
LGTM

PR #130, opened 2022-05-23
Cvsl 573 expose bookingid licencesummary

mgreen-mojhub:


andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceSummary.kt
```diff
@@ -73,4 +73,7 @@ data class LicenceSummary(
 
   @Schema(description = "The username of the responsible probation officer", example = "jsmith")
   val comUsername: String?,
+
+  @Schema(description = "The bookingId associated with the licence", example = "773722")
+  val bookingId: Long?,
```
andrewrlee, line 78: meh.. is this nullable?

mgreen-mojhub: I cannot see a scenario where we would have no bookingId. But I'm not 100% certain. Most of the fields are marked as nullable. I'm unsure if this is due to technical reasons, or it's just being defensive. Once I'm more familiar with service, I may be able to revisit the models and remove some of the nullable fields.

PR #131, opened 2022-05-25
CVSL-573

mgreen-mojhub:
Create new endpoint for updating licences to INACTIVE.
This compliments the existing activate licences endpoint, and is required for updating any licence where a HDC approved licence has been found for the same offender. In this scenario, the HDC licence invalidates the standard licence.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
@@ -541,6 +541,42 @@ class LicenceController(private val licenceService: LicenceService) {
     licenceService.activateLicences(request)
   }
 
+  @PostMapping(value = ["/inactivate-licences"])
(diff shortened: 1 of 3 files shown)


andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
```diff
@@ -541,6 +541,42 @@ class LicenceController(private val licenceService: LicenceService) {
...
+      ),
+      ApiResponse(
+        responseCode = "403",
+        description = "Forbidden, requires an appropriate role",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      )
+    ]
+  )
+  fun inactivateLicences(
+    @Valid @RequestBody request: List<Long>
```
andrewrlee, line 575: is there a way to annotate these in swagger so you can describe what they are in the docs?

PR #134, opened 2022-06-10
Cvsl 596 com update id changes

mgreen-mojhub:
When updating a COM, check for a match against username or staffId. Its possible the staffId can be changed in nDelius to link to another account. If this instance, we should update the existing record in CVL to reflect the new username and or staffId.

Failing to do so will result in a key constraint failure; causing a crash.

Check the COM details have changed before performing any update.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/CommunityOffenderManagerRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/CommunityOffenderManagerRepository.kt
@@ -7,5 +7,6 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.CommunityOff
 @Repository
 interface CommunityOffenderManagerRepository : JpaRepository<CommunityOffenderManager, Long> {
   fun findByStaffIdentifier(staffIdentifier: Long): CommunityOffenderManager?
+  fun findByStaffIdentifierOrUsername(staffIdentifier: Long, username: String): List<CommunityOffenderManager>?
   fun findByUsernameIgnoreCase(username: String): CommunityOffenderManager?
 }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/CommunityOffenderManagerRepository.kt
```diff
@@ -7,5 +7,6 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.CommunityOff
 @Repository
 interface CommunityOffenderManagerRepository : JpaRepository<CommunityOffenderManager, Long> {
   fun findByStaffIdentifier(staffIdentifier: Long): CommunityOffenderManager?
+  fun findByStaffIdentifierOrUsername(staffIdentifier: Long, username: String): List<CommunityOffenderManager>?
```
andrewrlee, line 10: think this will always be a non-null list - just empty if no matches

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -12,26 +13,64 @@ class ComService(
...
-        firstName = comDetails.firstName,
-        lastName = comDetails.lastName
+    }
+
+    if (comResult.count() > 1) {
+      log.warn(
+        "More then one COM record found for staffId {} username {}",
+        comDetails.staffIdentifier,
+        comDetails.staffUsername
+      )
```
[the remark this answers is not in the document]
mgreen-mojhub: Good question. We can receive "event bombs" where we have many event fired within a second or so. In this instance, the API can throw a lot of exceptions in quick succession.

When the DB was throwing exceptions on the key constraint, the app become very slow and unresponsive. Since exceptions are heavy to process, I opted for a log in place of an exception. But it could be changed to just return with no updates. 

I need to monitor for this scenario, as it would require Delius to consolidate the two accounts into one. Unless theres a legitimate reason for having more than one account with the same username and or staffId

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/CommunityOffenderManagerRepository.kt
mgreen-mojhub: Thanks for the info

PR #135, opened 2022-06-22
CVSL-603

mgreen-mojhub:
Check licence dates are not null before making any date checks between the licence entity and the update request

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
@@ -876,13 +876,33 @@ class LicenceService(
       updatedByUsername = username
     )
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -876,13 +876,33 @@ class LicenceService(
...
+
+    val ledChanged = (
+      sentenceDatesRequest.licenceExpiryDate?.let { it != licenceEntity?.licenceExpiryDate }
+        ?: licenceEntity?.licenceExpiryDate
+      ) == true
+
+    val sedChanged = (
+      sentenceDatesRequest.sentenceEndDate?.let { it != licenceEntity?.sentenceEndDate }
+        ?: licenceEntity?.sentenceEndDate
+      ) == true
```
andrewrlee, line 892: maybe extracting out a helper function for hasDateChange might help?

PR #156, opened 2022-10-06
CVSL-669 inactivate license for re-sentenced offender

dibyatanu:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/LicenceIntegrationTest.kt
```diff
@@ -680,7 +680,44 @@ class LicenceIntegrationTest : IntegrationTestBase() {
...
+  fun `Update sentence dates should set license status to inactive when the offender has a new future release date`() {
+    prisonApiMockServer.stubGetHdcLatest()
+
+    webTestClient.put()
+      .uri("/licence/id/3/sentence-dates")
+      .bodyValue(
+        UpdateSentenceDatesRequest(
+          conditionalReleaseDate = LocalDate.now().plusDays(5),
+          actualReleaseDate = LocalDate.now().plusDays(2),
+          sentenceStartDate = LocalDate.parse("2021-09-11"),
```
GurnankCheema, line 696: I'd be inclined to have any future dates as relative to LocalDate.now() like you've done on L695, just to prevent tests filing unexpectedly when the future dates have been reached.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -973,6 +975,9 @@ class LicenceService(
         }
     }
   }
+  private fun hasOffenderResentensedWithActiveLicense(sentenceDatesRequest: UpdateSentenceDatesRequest, licenceEntity: EntityLicence): Boolean {
+    return (sentenceDatesRequest.conditionalReleaseDate?.isAfter(LocalDate.now()) == true || sentenceDatesRequest.actualReleaseDate?.isAfter(LocalDate.now()) == true) && licenceEntity.statusCode == ACTIVE
+  }
```
mgreen-mojhub, line 980: Instead of an OR condition, check for actualReleaseDate being present. If so, use that value, otherwise use the conditionalReleaseDate. it's possible (but unlikely) the conditionalReleaseDate could be in the future, but the actualReleaseDate may be in the past. 

The actualReleaseDate is the definitive date, otherwise we rely on the conditionalReleaseDate.

I would possibly add a single check at the top of the function:
`IF (licenceEntity.statusCode !== ACTIVE`) return false`

To simplify the logic.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/LicenceIntegrationTest.kt
```diff
@@ -680,7 +680,44 @@ class LicenceIntegrationTest : IntegrationTestBase() {
...
+          sentenceEndDate = LocalDate.parse("2024-09-11"),
+          licenceStartDate = LocalDate.parse("2023-09-11"),
+          licenceExpiryDate = LocalDate.parse("2024-09-11"),
+          topupSupervisionStartDate = LocalDate.parse("2024-09-11"),
+          topupSupervisionExpiryDate = LocalDate.parse("2025-09-11"),
```
mgreen-mojhub, line 701: For future dates, use a relative now + time to prevent future issues when running the tests.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -1351,7 +1351,255 @@ class LicenceServiceTest {
...
+        sentenceEndDate = LocalDate.parse("2024-09-11"),
+        licenceStartDate = LocalDate.parse("2023-09-11"),
+        licenceExpiryDate = LocalDate.parse("2024-09-11"),
+        topupSupervisionStartDate = LocalDate.parse("2024-09-11"),
+        topupSupervisionExpiryDate = null,
+      )
+    )
+
+    val licenceCaptor = ArgumentCaptor.forClass(EntityLicence::class.java)
+    val auditCaptor = ArgumentCaptor.forClass(EntityAuditEvent::class.java)
```
andrewrlee, line 1380: there's some [nice syntax](https://github.com/mockito/mockito-kotlin/tree/a6f860461233ba92c7730dd42b0faf9ba2ce9281#argument-captors) for doing this in kotlin mockito.

Though looks like this is in keeping with how its done with other tests in this PR so probably not really worth changing

dibyatanu: Thank you Andrew for point it out.  Since I am new to Kotlin , I used existing test as a reference. But I will bear that in my for future

mgreen-mojhub (approved):

PR #158, opened 2022-10-10
CVSL-348

mgreen-mojhub:
Add endpoint for creating one additional condition to a licence 
Add endpoint for deleting one additional condition from a licence 
Refactor update additional conditions logic to handle multiple conditions with the same condition type


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -220,50 +222,113 @@ class LicenceService(
...
+        conditionText = request.conditionText,
+        expandedConditionText = request.expandedText,
+        conditionCategory = request.conditionCategory,
+        licence = licenceEntity,
+        conditionSequence = request.sequence
+      )
+    )
+
+    val updatedLicence = licenceEntity.copy(
+      additionalConditions = newConditions,
```
andrewrlee, line 255: I learnt recently that `.plus` or `+` works on immutable lists and returns a new copy:
so I think you could skip intermediate variable:
```
additionalConditions = additionalConditions + newCondition
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -220,50 +222,113 @@ class LicenceService(
...
+      dateLastUpdated = LocalDateTime.now(),
+      updatedByUsername = username,
+    )
+
+    licenceRepository.saveAndFlush(updatedLicence)
+
+    // return the newly added condition.
+    val newCondition = licenceEntity.additionalConditions.filter { it.conditionCode == request.conditionCode }.maxBy { it.id }
```
andrewrlee, line 263: slightly suprised, it clever enough to update the persistent entity. I'd have been tempted to just modify the existing licenceEntity instead. It's relying on mutation.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -220,50 +222,113 @@ class LicenceService(
...
+
+    val newAdditionalConditions =
+      submittedAdditionalConditions.filter { licenceEntity.additionalConditions.none { submittedCondition -> submittedCondition.conditionCode == it.conditionCode } }
+        .map { newAdditionalCondition -> newAdditionalCondition.copy(expandedConditionText = newAdditionalCondition.conditionText, licence = licenceEntity) }
+
+    val removedAdditionalConditionsList =
+      licenceEntity.additionalConditions.filter { submittedAdditionalConditions.none { submittedCondition -> submittedCondition.conditionCode == it.conditionCode } }
+
+    val updatedAdditionalConditions =
+      licenceEntity.additionalConditions.filter { condition -> removedAdditionalConditionsList.none { it.conditionCode == condition.conditionCode } }
```
andrewrlee, line 306: mmm.. could possibly add an `doesNotExistIn` helper extension function? 

```
licenceEntity.additionalConditions.filter { it.doesNotExistIn(removedAdditionalConditionsList) }
```

I think you'd be able to use it 3 times.. maybe the double negation in a filter is a bit brain blowing! 

Not super essential, just a bit difficult to read.

andrewrlee (approved):
LGTM

PR #160, opened 2022-10-18
CVSL-138 Adding change type 

andrewrlee:
Also starting to return info about conditions that have been removed in later version of the policy


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctionsTest.kt
```diff
@@ -0,0 +1,127 @@
...
+        code = "code-1", text = "some-text", data = listOf(
+          AdditionalConditionData(field = "ph-1")
+        )
+      )
+      assertThat(licencePolicyChanges(listOf(condition), listOf(policyCondition), placeHolders)).isEmpty()
+    }
+
+    @Disabled("Should pass?")
+    @Test
+    fun `if new fields added in policy then there should be a change?`() {
```
andrewrlee, line 56: Is this a bug?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctionsTest.kt
```diff
@@ -0,0 +1,132 @@
...
+          AdditionalConditionData(field = "ph-1")
+        )
+      )
+      val policyCondition = policyCondition().copy(code = "code-1", text = "some-text")
+
+      assertThat(licencePolicyChanges(listOf(condition), listOf(policyCondition), placeHolders)).isEmpty()
+    }
+
+    @Disabled("Should pass?")
+    @Test
```
andrewrlee, line 58: is this a bug? 

when the stored condition has no placeholders but the new policy version has some, it should return it as a change?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicencePolicyService.kt
```diff
@@ -17,11 +17,16 @@ data class ConditionChangesByType<T, U>(
...
+  DELETED, REPLACED, NEW_OPTIONS, TEXT_CHANGE
+}
+
 data class LicenceConditionChanges(
+  val changeType: ConditionChangeType,
   val code: String,
   val sequence: Int?,
   val previousText: String,
-  val currentText: String,
+  val currentText: String?,
```
andrewrlee, line 29: current text will be absent when the condition is deleted

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctions.kt
```diff
@@ -53,14 +57,26 @@ fun <T : ILicenceCondition> getPolicyPlaceholders(conditions: List<T>): HashMap<
...
+    val textHasChanged = match.text != condition.text
+    val dataChanges = condition.data.filter { d -> policyPlaceholders[condition.code]?.containsValue(d.field) == false }
+    val dataHasChanged = dataChanges.isNotEmpty()
+    when {
+      textHasChanged -> TEXT_CHANGE.of(condition, match.text, dataChanges)
+      dataHasChanged -> NEW_OPTIONS.of(condition, match.text, dataChanges)
+      else -> null
+    }
+  } else {
+    DELETED.of(condition, null, emptyList())
```
andrewrlee, line 73: need to distinguish between REPLACED and DELETED here

andrewrlee:
question around, why are the conditions that seem like they need to be deleted still present in the v2.1 config? 
e.g: [this](https://github.com/ministryofjustice/create-and-vary-a-licence/pull/387/files#diff-d9f6b308443c5b9c14f161b5c0fa02572ffcd4692260c287765ab5517ffc5941R20)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctions.kt
```diff
@@ -53,14 +57,26 @@ fun <T : ILicenceCondition> getPolicyPlaceholders(conditions: List<T>): HashMap<
...
+    val textHasChanged = match.text != condition.text
+    val dataChanges = condition.data.filter { d -> policyPlaceholders[condition.code]?.containsValue(d.field) == false }
+    val dataHasChanged = dataChanges.isNotEmpty()
+    when {
+      textHasChanged -> TEXT_CHANGE.of(condition, match.text, dataChanges)
+      dataHasChanged -> NEW_OPTIONS.of(condition, match.text, dataChanges)
+      else -> null
+    }
+  } else {
+    DELETED.of(condition)
```
andrewrlee, line 73: still need to distinguish between REPLACED and DELETED here
probably help to sort based on sequence too.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctionsTest.kt
SebNegahbanMoJ: For visibility:
We have discussed this on a call and the outcome was that we believe that this is correct expected behaviour, but will need to confirm. As there don't seem to be any changes of this type in the current policy changes (and with the imminent deadline), we're going to progress with the current behaviour and consider this post-release.

SebNegahbanMoJ (commented):
Approved with the caveat that I'm by no means fluent in Kotlin. The logic seems reasonable, but I'd ask for eyes from @mgreen-mojhub before giving it a green tick

PR #163, opened 2022-10-20
CVSL-138 licence versioning

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/PassInput.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy
+
+data class PassInput(
```
SebNegahbanMoJ, line 3: Reasonably minor, but should this class and filename be PSS rather than Pass?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicencePolicyService.kt
```diff
@@ -0,0 +1,64 @@
...
+  val code: String,
+  val sequence: Int?,
+  val previousText: String,
+  val currentText: String,
+  var dataChanges: List<AdditionalConditionData>
+)
+
+class LicencePolicyService(private val policies: List<LicencePolicy>) {
+
+  fun currentPolicy(): LicencePolicy = policies.maxBy { it.version }
```
SebNegahbanMoJ, line 30: Do we want to have a safeguard of locking this to v2.0 for now? I can put it in the front-end instead if you like?
Or are we just going to not release anything until Monday?

andrewrlee: probably not release till monday - should be easy to revert this if necessary

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/PassInput.kt
andrewrlee: possibly, we can look at fixing this

mgreen-mojhub (approved):
LGTM

PR #166, opened 2022-10-20
CVSL-138 Adding change type and condition replacements

andrewrlee:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/RealLicencePolicyServiceTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+  @Test
+  fun `Current policy version is returned`() {
+    val currentPolicy = licencePolicyService.policyByVersion("2.1")
+    val previousPolicy = licencePolicyService.policyByVersion("2.0")
+
+    val replacements = getSuggestedReplacements(previousPolicy, currentPolicy)
+
+    val diffs =
+      conditionChanges(previousPolicy.additionalConditions.ap, currentPolicy.additionalConditions.ap, replacements)
```
andrewrlee, line 26: this will print out the diffs between different licence versions

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctions.kt
```diff
@@ -25,42 +32,77 @@ fun <T : ILicenceCondition> conditionChanges(current: List<T>, other: List<T>) =
...
+  val conditionsToCheck = currentConditions.filter { current ->
+    presentConditions.any { present -> current.code == present.code }
+  }
+  val changes = conditionChanges(conditionsToCheck, previousConditions, allReplacements)
 
-  conditions.filter { it.requiresInput }.forEach {
-    val placeholders = placeholderPattern.findAll(it.text).iterator()
-    val variables = variablePattern.findAll(it.tpl ?: "").toSet().iterator()
-    val keyValuePairs = HashMap<String, String>()
+  return changes.map { change -> change.copy(sequence = presentConditions.find { condition -> condition.code == change.code }?.sequence) }
```
andrewrlee, line 46: this is kinda gnarly - re-stitch back licence specific info about the sequence.

But the tradeoff is that `conditionChanges` is simpler as we don't have to pass through licence specific data

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctions.kt
```diff
@@ -25,42 +32,77 @@ fun <T : ILicenceCondition> conditionChanges(current: List<T>, other: List<T>) =
...
+}
 
-    results[it.code] = keyValuePairs
+private fun <T : Any> conditionUpdated(
+  previous: IAdditionalCondition<T>,
+  current: IAdditionalCondition<T>,
+): LicenceConditionChanges? {
+  val textHasChanged = previous.text != current.text
+  val removed = (previous.inputs ?: emptyList()) - (current.inputs ?: emptyList()).toSet()
+  val added = (current.inputs ?: emptyList()) - (previous.inputs ?: emptyList()).toSet()
```
andrewrlee, line 67: just working out the difference between the two lists. Updates will appear in both lists - "You removed this variant of condition A and added this other variant of condition A"

This isn't used by the frontend but returning as it help with debugging and seeing why things are different

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctions.kt
```diff
@@ -25,42 +32,77 @@ fun <T : ILicenceCondition> conditionChanges(current: List<T>, other: List<T>) =
...
-    else null
+private fun conditionRemoved(
+  condition: ILicenceCondition,
+  allReplacements: List<Replacements>,
+): LicenceConditionChanges {
+  val replacements = allReplacements.find { it.code === condition.code }
+  return when (replacements?.changeType) {
+    DELETED -> DELETED.removal(condition, replacements.alternatives)
+    REPLACED -> REPLACED.removal(condition, replacements.alternatives)
+    else -> REMOVED_NO_REPLACEMENTS.removal(condition, emptyList())
```
andrewrlee, line 84: This isn't handled by the frontend and we don't have an example of this yet but it seemed quite an important thing to cover

mgreen-mojhub (approved):
Looks good

PR #167, opened 2022-10-21
CVSL-348

mgreen-mojhub:
Only remove conditions on update where the existing licence condition match the same type as the request.


andrewrlee (approved):
LGTM

PR #169, opened 2022-10-21
CVSL-138 Replacing old condition type with new one

SebNegahbanMoJ:
Preventing use of shared condition type definition, as the validations needed to diverge between the new and old versions


andrewrlee (approved):
LGTM

PR #170, opened 2022-10-22
CVSL-138 Versioning conditions

andrewrlee:
There isn't currently enough information stored in the DB to let us know if conditions that have changed between versions have been migrated.

We can general tell if there are text changes but not if the options have changed. (In V2.1 there are couple of examples where this has happened without a text change)

There is a migration included here that will set the condition version of all licences to be the same as the licence version.

When conditions are added or updated they will inherit the licences current version.

There was also an existing bug where the previous and next licences conditions were swapped around. I've added more test coverage for this.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctions.kt
```diff
@@ -32,18 +32,27 @@ fun <T : ILicenceCondition> conditionChanges(current: List<T>, other: List<T>) =
...
+  licence: Licence,
   previousConditions: List<IAdditionalCondition<T>>,
   currentConditions: List<IAdditionalCondition<T>>,
   allReplacements: List<Replacements>
 ): List<LicenceConditionChanges> {
-  val conditionsToCheck = currentConditions.filter { current ->
-    presentConditions.any { present -> current.code == present.code }
+  val conditionsToCheck = previousConditions.filter { policyCondition ->
+    licence.additionalLicenceConditions.any { licenceCondition ->
+      licenceCondition.code == policyCondition.code && licence.version != licenceCondition.version
```
andrewrlee, line 46: we now only consider conditions that have not been migrated

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyFunctions.kt
```diff
@@ -32,18 +32,27 @@ fun <T : ILicenceCondition> conditionChanges(current: List<T>, other: List<T>) =
...
+    }
   }
-  val changes = conditionChanges(conditionsToCheck, previousConditions, allReplacements)
+  val changes = conditionChanges(conditionsToCheck, currentConditions, allReplacements)
```
andrewrlee, line 49: previous and current conditions were swapped around previously, I hadn't got test coverage here so have added some

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/VersionDifferencesTest.kt
```diff
@@ -11,17 +11,44 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.ConditionCh
...
+    val currentPolicy = licencePolicyService.policyByVersion("2.1")
+    val previousPolicy = licencePolicyService.policyByVersion("1.0")
+
+    val replacements = getSuggestedReplacements(previousPolicy, currentPolicy)
+
+    val diffs =
+      conditionChanges(previousPolicy.additionalConditions.ap, currentPolicy.additionalConditions.ap, replacements)
+
+    showDiffs(diffs)
+  }
```
andrewrlee, line 48: There are quite a lot of records in the DB that are at version 1.0 that could be migrated to version 2.1. 

I think it doesn't matter as there are no backward incompatible changes so the diffs is empty

src/test/resources/test_data/seed-licence-summaries.sql
```diff
@@ -1,9 +1,10 @@
...
   com_staff_id,
   type_code,
   status_code
 ) values
-(1,1,'AP','IN_PROGRESS'),
-(2,1,'AP','APPROVED'),
-(3,1,'AP','REJECTED');
+(1,'1.0',1,'AP','IN_PROGRESS'),
+(2,'1.0',1,'AP','APPROVED'),
+(3,'1.0'',1,'AP','REJECTED');
```
SebNegahbanMoJ, line 10: Is the second apostrophe in `'1.0''` intentional?

andrewrlee: this wasn't being used. deleted the file

mgreen-mojhub (approved):

PR #185, opened 2022-11-18
Re-adding the ability to select options for condition 4a, but without the course centre input

SebNegahbanMoJ:
Changing the behaviour of condition 4a to match expectation of users


mgreen-mojhub (approved):
LGTM

PR #188, opened 2022-12-05
CVSL-738-licence-override-service

mgreen-mojhub:
Licence override service with initial functionality to change allocated status code for a licence. Only allows one status to be used against each CRN number, unless making a licence INACTIVE.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceOverrideController.kt
```diff
@@ -0,0 +1,72 @@
...
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      )
+    ]
+  )
+  fun changeStatus(
+    @PathVariable("licenceId") licenceId: Long,
+    @RequestBody @Valid request: OverrideLicenceStatusRequest
+  ) {
+    val licence = licenceOverrideService.getLicenceById(licenceId)
+    licenceOverrideService.changeStatus(licence!!, request.statusCode, request.reason)
```
GurnankCheema, line 70: calling licenceOverrideService twice. Do as one step in the service?

mgreen-mojhub: It could be within the same method. Having two methods ensures a separation of concerns, but requires two calls on the controller side (or a call to the method within the same service). Its largely subjective to a point. @andrewrlee any preference in this instance? 

The existing LicenceService is huge and often has the same code copied and pasted within each method to get the Licence.

GurnankCheema (dismissed):
LGTM, minor comment

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceOverrideController.kt
andrewrlee: preference on one method call - then you don't need to expose a public `getLicenceById` method on the override service but not a big deal

A massive plus one on having a separate service!

PR #189, opened 2022-12-05
CVSL-740

mgreen-mojhub:
Include JBIG2 plugin for pdfbox in order to support PDF files with embedded images using this format. Minor bump to postgresql driver.


andrewrlee (approved):
LGTM

PR #192, opened 2022-12-19
Adding additional conditions type to v1.0 policy to fix broken validations

SebNegahbanMoJ:
Validations are currently broken on v1.0 licences as the types are missing from the policy doc.
This also causes any conditions with address fields to be unsubmittable because the lack of a `stringify` method it causes a 400 error (introduced by the validator) means the address submits in a format that the API cannot handle.


mgreen-mojhub (approved):
LGTM

PR #193, opened 2022-12-19
CVSL-750 fixing N+1 issue

andrewrlee:
Before the change, "Get licences matches - no filters" test used to generate the following queries:

```
Hibernate: select ... from licence licence0_
Hibernate: select ... from community_offender_manager communityo0_ where communityo0_.id=? Hibernate: select ... from community_offender_manager communityo0_ where communityo0_.id=? Hibernate: select ... from community_offender_manager communityo0_ where communityo0_.id=? ...
```

After this change it generates:
```
select ... from licence licence0_ inner join community_offender_manager communityo1_ on licence0_.responsible_com_id=communityo1_.id
```
Combined with another bug this would generate up to an extra 25-30k sql queries per minute

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
@@ -30,6 +30,10 @@ fun LicenceQueryObject.toSpecification(): Specification<Licence> = and(
   hasPdusIn(pdus),
   hasProbationAreaCodeIn(probationAreaCodes)
 )
+  .and { root, query, _ ->
+    root.fetch<Licence, CommunityOffenderManager>("responsibleCom", JoinType.INNER)
+    query.restriction
+  }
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
```diff
@@ -30,6 +30,10 @@ fun LicenceQueryObject.toSpecification(): Specification<Licence> = and(
   hasPdusIn(pdus),
   hasProbationAreaCodeIn(probationAreaCodes)
 )
+  .and { root, query, _ ->
+  root.fetch<Licence, CommunityOffenderManager>("responsibleCom", JoinType.INNER)
+  query.restriction
+}
```
andrewrlee, line 36: There was an inner join on the `hasResponsibleComIn` predicate below but this would only added in if the com id was passed in. 

It also didn't seem work, I think due to laziness

mgreen-mojhub (approved):
LGTM

PR #200, opened 2023-02-03
CVL-626 Moving to defining licence policies as code

andrewrlee:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PolicyVersionDifferencesTest.kt
```diff
@@ -61,6 +62,24 @@ class PolicyVersionDifferencesTest() {
     showDiffs(diffs)
   }
 
+  @Test
+  fun checkCodeVsJson() {
+    assertThat(licencePolicyService.policyByVersion("1.0"))
+      .usingRecursiveComparison()
+      .withStrictTypeChecking()
+      .isEqualTo(POLICY_V1_0)
```
andrewrlee, line 70: these tests prove the json and kotlin versions are equivalent and will ensure that the json and kotlin representation stay in sync

andrewrlee: (the recursive comparison bit just gives better feedback of the specific level where difference occur,  the strict typechecking is that it does instance checking too)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies/policyV1.kt
```diff
@@ -0,0 +1,1506 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
```
mgreen-mojhub, line 1: Small point, but package could / maybe should be service.policies

andrewrlee: yes, totally agree. Is it alright to change in a follow up?

mgreen-mojhub: No problem, nothing urgent

SebNegahbanMoJ (approved):

PR #201, opened 2023-02-03
DCS-626 defaulting nullable fields

andrewrlee:
This just means that the policy files don't have to specify nulls so knock off ~1800 lines of code


mgreen-mojhub (approved):
LGTM

PR #203, opened 2023-02-08
Duplicating 'unapproved licence' email to CVL inbox

SebNegahbanMoJ:
Refs: [CVSL-779](https://dsdmoj.atlassian.net/browse/CVSL-779)

[CVSL-779]: https://dsdmoj.atlassian.net/browse/CVSL-779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/NotifyService.kt
```diff
@@ -91,18 +92,28 @@ class NotifyService(
...
       )
       log.info("Notification sent to ${it.comEmail} informing edited licence not reapproved by CRD for prisoner ${it.forename} ${it.surname}")
+
+      sendEmail(
+        unapprovedLicenceByCrdTemplateId,
+        internalEmailAddress,
+        offenderDetails,
+        null
+      )
+      log.error("Error sending email to internal email address")
```
andrewrlee, line 116: presumably we don't want this log line anymore?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/NotifyServiceTest.kt
```diff
@@ -233,6 +234,7 @@ class NotifyServiceTest {
       variationForReApprovalTemplateId = TEMPLATE_ID,
       unapprovedLicenceByCrdTemplateId = TEMPLATE_ID,
       client = notificationClient,
+      internalEmailAddress = "testemail@probation.gov.uk"
```
andrewrlee, line 237: probably want to extend the `send unapproved licence email to probation practitioner` to ensure you are sending to the new email address

SebNegahbanMoJ: Ah did you mean replace it with the `EMAIL_ADDRESS` constant that I had missed previously? 😅

andrewrlee: That test is presumably only verifying that its sending emails to the PP. Should also now add checks that we're sending email to the internal email address?

SebNegahbanMoJ: Oh I see! Gotcha, yes I will do.
Also I'm having one of 'those' mornings, of course I don't want to use the existing `EMAIL_ADDRESS`, the whole point is that it's a different one 🤦

andrewrlee (approved):
LGTM

PR #204, opened 2023-02-08
CVL-626 defaulting nullable fields

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionAp.kt
```diff
@@ -4,10 +4,10 @@ data class AdditionalConditionAp(
   override var code: String,
   override val category: String,
   override val text: String,
-  override val tpl: String?,
```
andrewrlee, line 7: this just means that you don't have to specify `blah = null` when creating new instances

andrewrlee: which massively shortens the policy files below

SebNegahbanMoJ (approved):

SebNegahbanMoJ (approved):

justin-gardiner (approved):
LGTM

PR #208, opened 2023-04-17
CVSL-863: Add spring multipart properties to override default max fil…e upload size. Currently set to 10MB maximum file for now, default was 1MB.

justin-gardiner:


andrewrlee (approved):
LGTM

PR #211, opened 2023-04-21
CVSL-863: Update policy labels for exclusion zone map.

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies/policyV1.kt
```diff
@@ -495,12 +495,12 @@ val POLICY_V1_0 = LicencePolicy(
...
         inputs = listOf(
           Input(
-            label = "Enter the name of area shown on the map",
+            label = "Enter the name of the area shown on the map",
             name = "outOfBoundArea",
             type = "text",
           ),
           Input(
-            label = """Find and select the <a class="govuk-link" href="https://mapmaker.field-dynamics.co.uk/moj/map/default" rel="noreferrer noopener" target="_blank">Mapmaker PDF map</a> to include on the licence""",
+            label = """Select a PDF map of the area this person must not enter""",
```
SebNegahbanMoJ, line 503: There are a few labels that have three sets of quotes, is that intentional?
Not your change obviously, I just can't figure out why it would be like this. Is there some CSS spacing magic happening because it's creating empty <p> tags or something?

justin-gardiner: triple quote allows multi-line strings, so it was needed before but not now. I have updated to single quote

SebNegahbanMoJ (approved):

PR #216, opened 2023-05-09
CVSL-758 - Forcing COM usernames to be uppercase and removing case sensitivity on lookup

SebNegahbanMoJ:
It appears that duplicate COM accounts were being created due to a combination of a few factors:
- HMPPS auth returns a normalised version of the username, in uppercase, whereas Delius returns the username as it exists (often lowercase)
- When a COM gets given a new account (possible due to moving probation region), they keep the same username but get a new Staff Indentifier
- Our existing COM record lookup is case sensitive

These three issues combine into a situation where it's possible for two accounts with the same username in different cases to be presumed by CVL to be two unique users, as they have different staff IDs and different casing of username. This then later causes an error when a subsequent search finds both COM records connected to one (differently-cased) username.

This fix adds `IgnoreCase` to the CVL lookup on update, meaning the new Staff Identifier will simply override the existing account info. It also begins a standard of making all CVL COM usernames uppercase, for data consistency.

This PR should be accompanied by a migration to update all existing COM usernames to be uppercase.


justin-gardiner:
will we need to clean up existing duplicates before this goes out?

SebNegahbanMoJ:
> will we need to clean up existing duplicates before this goes out?

Tentative yes in the sense that this will cause issues for users with duplicates, but also the users with duplicates will already be having issues, so it probably won't change much!
That said, I'm going to be writing a migration to update all of the existing lowercase usernames to uppercase for data consistency anyway, and that would likely fail with duplicates present, so they'll need to be resolved for that regardless.

justin-gardiner (approved):
LGTM

PR #222, opened 2023-05-18
CVSL-1030 Add in community api client and probation search client in api

gurps-singh:
This PR is to add both community API and probation search clients to the API to allow us to call endpoints on the community API and probation search from the backend.


andrewrlee (approved):
LGTM

PR #223, opened 2023-05-18
CVSL-1030 Add call to community api endpoint to get team codes for a user

gurps-singh:
This PR is to add an endpoint call to the community api to retrieve the team codes for a user based on their staff identifier.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/CommunityApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/CommunityApiClient.kt
@@ -32,7 +32,7 @@ class CommunityApiClient(@Qualifier("oauthCommunityApiClient") val communityApiC
     } catch (exception: WebClientResponseException) {
       with(exception) {
         val uriPath = request?.uri?.path
-        log.error("No user found for staff when calling the community-api $uriPath with the following message ${exception.message}")
+        log.error("No user found for staff identifier when calling the community-api $uriPath with the following message ${exception.message}")
       }
       return emptyList()
     }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/CommunityApiClient.kt
```diff
@@ -0,0 +1,40 @@
...
+        .uri("/secure/staff/staffIdentifier/$staffIdentifier")
+        .accept(MediaType.APPLICATION_JSON)
+        .retrieve()
+        .bodyToMono(User::class.java)
+        .block()
+      if (user === null) {
+        return emptyList()
+      } else {
+        val userTeams = user.teams
+        return userTeams.map { team: Team -> team.code }
```
andrewrlee, line 30: super minor and not worth changing, probably more idiomatic kotlin to use `it`.

I'd personally remove the temp variables and just `user.teams.map { it.code }`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/CommunityApiClient.kt
```diff
@@ -0,0 +1,40 @@
...
+  fun getTeamsCodesForUser(staffIdentifier: Long): List<String> {
+    try {
+      val user = communityApiClient
+        .get()
+        .uri("/secure/staff/staffIdentifier/$staffIdentifier")
+        .accept(MediaType.APPLICATION_JSON)
+        .retrieve()
+        .bodyToMono(User::class.java)
+        .block()
+      if (user === null) {
```
andrewrlee, line 26: this should never be null and if it was then throwing an exception would be a fair enough response rather than failing silently. 

So few options:  

could `!!` - you'd get a null pointer referencing that line of code but not a lot of context in the message 

could ` ?: throw IllegalStateException("some context")

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

PR #224, opened 2023-05-22
Fix env var name

andrewrlee:
was expecting no underscore if variable name


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/config/WebClientConfiguration.kt
```diff
@@ -18,7 +18,7 @@ class WebClientConfiguration(
   @Value("\${hmpps.prison.api.url}") private val prisonApiUrl: String,
   @Value("\${hmpps.prisonregister.api.url}") private val prisonRegisterApiUrl: String,
   @Value("\${hmpps.community.api.url}") private val communityApiUrl: String,
-  @Value("\${hmpps.probationSearch.api.url}") private val probationSearchApiUrl: String,
+  @Value("\${hmpps.probationsearch.api.url}") private val probationSearchApiUrl: String,
```
andrewrlee, line 21: this doesn't really make a distance as the appropriate env var name is derived by upper casing and replacing `.` with `_`, but just to be consistent with the prisonregister example above

SebNegahbanMoJ (approved):

PR #228, opened 2023-05-30
Upgrading orb version, hoping to fix intermittent deployment issue 

andrewrlee:
Potentially fix app version disappearing during deployment

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/config/CacheConfiguration.kt
justin-gardiner, line 1: Did you mean to add this?

PR #232, opened 2023-06-05
CVSL-1044 Removing unused mailing list

andrewrlee:


src/main/resources/migration/common/V23__delete_mailing_list.sql
justin-gardiner, line 1: I added a V23 script already, V23__licence_edit_of_id_column.sql

SebNegahbanMoJ (approved):

PR #235, opened 2023-06-07
CSVSL-1048 Improving performance by reducing the number of n+1 issues

andrewrlee:
Depending on the way the criteria object was created it could trigger up to 4*n+1 queries as it would eagerly load the various collections of conditions and com details. This was all unnecessary as the matches endpoint produces LicenceSummarys which do not surface any info about additional conditions and only provide info about the responsible com

There still seems to be an n+1 query as it seems to retrieve the created com even though its not being used - still not really sure why!

None of the other methods retrieving multiple Licences require access to associated entities (conditions/other coms). If we need to add these methods we should add custom methods that have appropriate fetch strategies. Using the generic "match" endpoint would be inappropriate for that

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/CommunityOffenderManager.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/CommunityOffenderManager.kt
@@ -5,7 +5,6 @@ import jakarta.persistence.Entity
 import jakarta.persistence.GeneratedValue
 import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
-import jakarta.persistence.OneToMany
 import jakarta.persistence.Table
 import jakarta.validation.constraints.NotNull
 import java.time.LocalDateTime
@@ -29,15 +28,6 @@ data class CommunityOffenderManager(
   val firstName: String?,
 
   val lastName: String?,
-
-  @OneToMany(mappedBy = "responsibleCom")
-  val licencesResponsibleFor: List<Licence> = emptyList(),
-
-  @OneToMany(mappedBy = "createdBy")
-  val licencesCreated: List<Licence> = emptyList(),
-
-  @OneToMany(mappedBy = "submittedBy")
-  val licencesSubmitted: List<Licence> = emptyList(),
 
   val lastUpdatedTimestamp: LocalDateTime? = null,
 )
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/LicenceMatchingIntegrationTest.kt
```diff
@@ -13,6 +13,24 @@ import java.time.LocalDate
 
 class LicenceMatchingIntegrationTest : IntegrationTestBase() {
 
+  @Test
+  @Sql(
+    "classpath:test_data/seed-a-few-licences.sql"
+  )
+  fun `Find approved licences`() {
```
andrewrlee, line 20: Improvements for each test in this suite: 

`Find approved licences`: 21 queries -> 5 queries (4n + 1 -> n + 1)
  
`Get licences matches - no filters`: 4 queries-> 1 query

`Get licences matches - by list of staff identifiers`:   8 queries -> 2 queries
  
  `Get licence matches - by list of staff identifiers and statuses`: 5 queries -> 2 queries

 
  `Get licence matches - by list of prisons and statuses`: 8 -> 2 queries
  
`Get licence matches - no matching filters`: 1 query  -> 1 query
  
 `Get licence matches - sort by conditional release date`: 8 queries -> 2 queries

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -119,6 +119,7 @@ class LicenceService(
     return createLicenceResponse
   }
 
+  @Transactional
```
andrewrlee, line 122: this is required as there needs to be a session to lazily load the related entities when building up the detailed DTO response

Previously it was relying on retrieving everything through eager loading

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/CommunityOffenderManager.kt
```diff
@@ -30,14 +29,5 @@ data class CommunityOffenderManager(
...
 
-  @OneToMany(mappedBy = "responsibleCom")
-  val licencesResponsibleFor: List<Licence> = emptyList(),
-
-  @OneToMany(mappedBy = "createdBy")
-  val licencesCreated: List<Licence> = emptyList(),
-
-  @OneToMany(mappedBy = "submittedBy")
-  val licencesSubmitted: List<Licence> = emptyList(),
-
```
andrewrlee, line 41: These backward relations aren't currently used so removed (can always add in if necessary)

justin-gardiner (approved):
lgtm

justin-gardiner (approved):
lgtm

PR #237, opened 2023-06-07
CVSL-1023 Moving licence activation job to API

SebNegahbanMoJ:
Currently merging into the branch that adds the required API endpoints. Can be changed to main after that PR is merged.

Moves the licence activation job to the API and adds new logic for activating IS91 licences


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
```diff
@@ -713,7 +713,7 @@ class LicenceController(
   fun activateLicences(
     @Valid @RequestBody request: List<Long>
   ) {
-    licenceService.activateLicences(request)
+    licenceService.activateLicencesByIds(request)
```
andrewrlee, line 716: are we going to be able to get rid of these after moving over to the new endpoint?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -429,10 +427,16 @@ class LicenceService(
   }
 
   @Transactional
-  fun inActivateLicences(licenceIds: List<Long>) {
+  fun activateLicencesByIds(licenceIds: List<Long>) {
```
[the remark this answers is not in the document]
andrewrlee: It's not strictly necessary in this case but you shouldn't shy away from it. 

sure it can be overused but it's idiomatic to use it. 

You have all the context you need from the method signature which is available everywhere

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
andrewrlee: if so you could add a @Deprecated annotation to methods that can be deleted. This will grey them out of swagger docs and add warnings in intelliJ if you use them

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -429,10 +427,16 @@ class LicenceService(
...
     val matchingLicences =
-      licenceRepository.findAllById(licenceIds).filter { licence -> licence.statusCode == APPROVED }
-    val inActivatedLicences = matchingLicences.map { licence -> licence.copy(statusCode = INACTIVE) }
+      licenceRepository.findAllById(licenceIds).filter { it.statusCode == APPROVED }
+    if (matchingLicences.any()) {
```
andrewrlee, line 433: No point checking here, the lower level will ensure what to do if there are no licences.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationService.kt
```diff
@@ -0,0 +1,94 @@
...
+  private val prisonerSearchApiClient: PrisonerSearchApiClient
+) {
+
+  @Transactional
+  fun licenceActivationJob() {
+    val potentialLicences = licenceRepository.getApprovedLicencesOnOrPassedReleaseDate()
+    val (eligibleLicences, ineligibleLicences) = determineLicenceEligibility(potentialLicences)
+    val licencesToActivate = determineEligibleLicencesToActivate(eligibleLicences)
+
+    if (licencesToActivate.any()) {
```
andrewrlee, line 27: No point checking here, the lower level will safely handle what to do if there are no licences.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
SebNegahbanMoJ: We may be able to get rid of the activation method, but there's a different process that hits the deactivation endpoint (the licence expiry job for when licences reach their end dates).
Truthfully I was leaving it for now until the updated job is working, just in case something goes wrong, because then all we have to do is rollback the front-end changes and the old job will still work.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
SebNegahbanMoJ: True, it just prevents an unnecessary method call if there's nothing that matches.
I can remove it if you'd prefer though!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/PrisonApiClient.kt
```diff
@@ -27,6 +28,41 @@ class PrisonApiClient(@Qualifier("oauthPrisonClient") val prisonerApiWebClient:
...
+      .uri("/bookings/offence-history")
+      .bodyValue(bookingIds)
+      .accept(MediaType.APPLICATION_JSON)
+      .retrieve()
+      .bodyToMono(typeReference<List<PrisonerOffenceHistory>>())
+      .onErrorResume { webClientErrorHandler(it) }
+      .block() ?: emptyList<PrisonerOffenceHistory>()
+  }
+
+  fun getIS91AndExtraditionBookingIds(bookingIds: List<Long>): List<Long> {
```
andrewrlee, line 55: this is a bit of business code and probably belongs in the service layer with some tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/PrisonApiClient.kt
```diff
@@ -27,6 +28,41 @@ class PrisonApiClient(@Qualifier("oauthPrisonClient") val prisonerApiWebClient:
...
+    val allOffenceHistories = offenceHistories(bookingIds)
+    val desiredResultCodes: Array<String> = arrayOf("5500", "4022", "3006", "5502")
+    val is91AndExtraditionOffenceHistories = allOffenceHistories.filter {
+      desiredResultCodes.contains(it.primaryResultCode) ||
+        desiredResultCodes.contains(it.secondaryResultCode) ||
+        it.offenceCode == "IA99000-001N"
```
andrewrlee, line 61: this could be a constant too

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationService.kt
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private fun Licence.standardLicenceForActivation(prisoners: List<PrisonerSearchPrisoner>): Boolean {
+    val prisoner: PrisonerSearchPrisoner = prisoners.first {
+      it.prisonerNumber == this.nomsId
+    }
+    if (this.actualReleaseDate != null &&
+      this.actualReleaseDate <= LocalDate.now() &&
+      prisoner.status?.startsWith("INACTIVE") == true
+    ) {
```
andrewrlee, line 89: ` if (a) { return true } else { return false } ` => `return a`

SebNegahbanMoJ: If I understand correctly, you mean I should just drop the `== true`?
The potential null-state of `prisoner.status` meant that doing `prisoner.status?.startsWith("INACTIVE")` without an explicit `true` check threw an error. Specifically a type mismatch as the condition was expecting a boolean but could receive a `null`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationService.kt
```diff
@@ -0,0 +1,94 @@
...
+      this.actualReleaseDate
+    } else {
+      this.conditionalReleaseDate
+    }
+
+    return releaseDate <= LocalDate.now()
+  }
+
+  private fun Licence.standardLicenceForActivation(prisoners: List<PrisonerSearchPrisoner>): Boolean {
+    val prisoner: PrisonerSearchPrisoner = prisoners.first {
```
andrewrlee, line 83: would `firstOrNull` be better here?
`first` will throw no such element exception if empty

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/PrisonApiClient.kt
```diff
@@ -27,6 +28,41 @@ class PrisonApiClient(@Qualifier("oauthPrisonClient") val prisonerApiWebClient:
       .onErrorResume { webClientErrorHandler(it) }
   }
 
+  fun hdcStatuses(bookingIds: List<Long>): List<PrisonerHdcStatus> {
```
andrewrlee, line 31: I know you're following the existing example but generally you prefix these with `get`, if you're just retrieving something

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationServiceTest.kt
```diff
@@ -0,0 +1,304 @@
...
+  @BeforeEach
+  fun reset() {
+    val authentication = mock<Authentication>()
+    val securityContext = mock<SecurityContext>()
+
+    whenever(authentication.name).thenReturn("smills")
+    whenever(securityContext.authentication).thenReturn(authentication)
+    SecurityContextHolder.setContext(securityContext)
+
+    org.mockito.kotlin.reset(
```
andrewrlee, line 47: would generally statically import these. magic key combo should allow you to do this.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -755,15 +755,33 @@ class LicenceServiceTest {
...
+  }
+
+  @Test
+  fun `activateLicencesByIds filters unapproved licences out before calling activateLicences`() {
+    whenever(licenceRepository.findAllById(listOf(1))).thenReturn(listOf(aLicenceEntity))
+
+    service.activateLicencesByIds(listOf(1))
+
+    verify(licenceRepository, times(1)).findAllById(listOf(1L))
+    verify(service, times(0)).activateLicences(anyList())
```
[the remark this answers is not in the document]
SebNegahbanMoJ: I'm happy to remove the checking for empty lists, but I'd probably still want to verify here that `activateLicences` had been called with an empty list rather than not being called with any list
ie I'd change it to `verify(service, times(1)).activateLicences(emptyList())`

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationServiceTest.kt
```diff
@@ -0,0 +1,304 @@
...
+      listOf(
+        prisonerSearchPrisoner
+      )
+    )
+    whenever(prisonApiClient.getIS91AndExtraditionBookingIds(listOf(54321))).thenReturn(emptyList())
+    whenever(prisonApiClient.hdcStatuses(listOf(54321))).thenReturn(emptyList())
+
+    service.licenceActivationJob()
+
+    verify(licenceService, times(0)).activateLicences(anyList())
```
andrewrlee, line 196: rather than using times(0), I think you can pass `never()` as second param, or use `verifyNoMoreInteractions(licenceService)`.

but if you remove the unnecessary empty check then this could be verify passed an `emptyList()`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
andrewrlee: That's fine - good to add deprecated if it's your intention to remove the active one. Rollback would revert the annotation but it signifies to people "do not use it" and "this will go"

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
andrewrlee: think it'd make sense to, makes the code a bit simpler, and reduces need for extra tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationService.kt
andrewrlee: I meant the whole if/else  - 

you can just:
```
return this.actualReleaseDate != null 
  && this.actualReleaseDate <= LocalDate.now() 
  && prisoner.status?.startsWith("INACTIVE") == true

```

SebNegahbanMoJ: Oh yeah of course, I didn't even think about that!

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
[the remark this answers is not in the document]
andrewrlee: yep, that makes sense! I didn't understand that test, 

maybe you make it more explicit by not relying on a default value declared else where?:
```  
    val licence = aLicenceEntity.copy(statusCode = LicenceStatus.UNAPPROVED)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/PrisonApiClient.kt
```diff
@@ -27,6 +28,41 @@ class PrisonApiClient(@Qualifier("oauthPrisonClient") val prisonerApiWebClient:
...
+      .accept(MediaType.APPLICATION_JSON)
+      .retrieve()
+      .bodyToMono(typeReference<List<PrisonerOffenceHistory>>())
+      .onErrorResume { webClientErrorHandler(it) }
+      .block() ?: emptyList<PrisonerOffenceHistory>()
+  }
+
+  fun getIS91AndExtraditionBookingIds(bookingIds: List<Long>): List<Long> {
+    val allOffenceHistories = offenceHistories(bookingIds)
+    val desiredResultCodes: Array<String> = arrayOf("5500", "4022", "3006", "5502")
```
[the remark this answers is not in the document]
SebNegahbanMoJ: Unless I'm missing something, I don't think I'm able to explicitly make a `Set` a `const` as it's not a primitive type or String. I've made it a `val` and put it in a separate object of constants in the hopes that that's sufficient, but if there's a better way to handle it, please let me know!

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/IS91DeterminationServiceTest.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.jupiter.api.Test
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.reset
+import org.mockito.kotlin.whenever
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonApiClient
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonerOffenceHistory
+
+class IS91DeterminationServiceTest {
+  private val prisonApiClient = mock<PrisonApiClient>()
+  private val service = IS91DeterminationService(prisonApiClient)
```
andrewrlee, line 13: meh.. benefit of having this as a dependency is that you don't necessarily need to test its functionality here and can mock it out. 

You can just write tests against its contract, what happens if it filters bookings in or out - otherwise you're testing the same thing at multiple levels.

Probably not worth changing but worth keeping in mind

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/IS91DeterminationService.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonApiClient
+
+@Service
+class IS91DeterminationService(
+  private val prisonApiClient: PrisonApiClient
+) {
+
+  private object IS91Constants {
```
andrewrlee, line 11: meh, bit more idiomatic to make this a [companion object](https://kotlinlang.org/docs/object-declarations.html#companion-objects) now it sits inside a class

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/PrisonApiClient.kt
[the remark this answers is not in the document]
andrewrlee: yes- consts is a compile time constant. 

A static val of an immutable collection can be considered a constant, it's just not resolved at compile time. 
The arrayOf method  is called at runtime

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/IS91DeterminationService.kt
```diff
@@ -0,0 +1,25 @@
...
+    val resultCodes = setOf("5500", "4022", "3006", "5502")
+    const val offenceCode = "IA99000-001N"
+  }
+
+  fun getIS91AndExtraditionBookingIds(bookingIds: List<Long>): List<Long> {
```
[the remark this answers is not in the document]
SebNegahbanMoJ: I considered making it `filter` but as it filters based off a call inside the method to the prison API, I thought leaving it as `get` may be better to signify that it's making an external API call.
I can change it to something `filter`y if you think it would be clearer though!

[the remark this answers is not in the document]
andrewrlee: Maybe it doing a request to an external api is an implementation detail. 
I'm not sure what a good `filter`y name would be so lets leave it!

PR #240, opened 2023-06-14
Fixing prisoner search request structure and removing error handling for loud failures

SebNegahbanMoJ:
Removing error handling from the prisonerSearchApi call to prevent an issue where the licence activation job would seemingly complete, but have just failed quietly.

Also fixing the incorrect request formatting that was causing the job to fail.


andrewrlee (approved):
LGTM... though lack of test change is slightly disconcerting

PR #244, opened 2023-06-15
Cvsl 465 dont allow pp to vary licence conditions if led less than current date less than equals tused

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/TransformFunctions.kt
```diff
@@ -157,9 +158,20 @@ fun transform(licence: EntityLicence): ModelLicence {
...
 
+fun checkLicenceIsInPssPeriod(licence: EntityLicence): Boolean {
+  val led = licence.licenceExpiryDate
+  val tused = licence.topupSupervisionExpiryDate
+
+  if (led != null && tused != null) {
+    return led.isBefore(LocalDate.now()) && !(tused.isBefore(LocalDate.now()))
+  }
+  return false
+}
```
andrewrlee, line 173: maybe this could live on the EntityLicence as an `isInPssPeriod` function? - I find it's generally good to push code down layers as it tends to become easier to test. 

Then here you would just need to copy across the field.

this would be nice as it means we have business logic closer to the domain entity.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -139,6 +139,22 @@ class LicenceServiceTest {
     assertThat(licence.nomsId).isEqualTo(aLicenceEntity.nomsId)
     assertThat(licence.bookingId).isEqualTo(aLicenceEntity.bookingId)
     assertThat(licence.pnc).isEqualTo(aLicenceEntity.pnc)
+    assertThat(licence.isInPssPeriod).isFalse()
+
+    verify(licenceRepository, times(1)).findById(1L)
+  }
+
+  @Test
+  fun `service transforms key fields of a licence in PSS period object correctly`() {
```
andrewrlee, line 148: You're LicenceEntityTest would then not need to worry about calling a service or calling a repository - you can get straight to the nitty gritty of testing the logic itself. 

It would also be good to test each branch in the conditions in the code 
```
  if (led != null && tused != null) {
    return led.isBefore(LocalDate.now()) && !(tused.isBefore(LocalDate.now()))
  }
  return false
```

what if led is null?
what if tused is null?
what if both are null?
what if led.isBefore now only 
what if tused.isBefore now only 
what if both are before now?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -1260,5 +1276,55 @@ class LicenceServiceTest {
       bookingId = 54321,
       dateCreated = LocalDateTime.of(2022, 7, 27, 15, 0, 0)
     )
+
+    val aPssLicenceEntity = EntityLicence(
```
andrewrlee, line 1280: if you need to create different variations of test data you can use the copy fun that is added to all data classes.
This is nice as you can really draw attention to what you are testing in each test and leave the rest of the fields with a default value.

```kt
val dto = SomeReallyComplicatedDataClassWith1000fields(....)

@Test
fun `check what makes SomeReallyComplicatedDataClassWith1000fields happy`() {
   val value = dto.copy(hasChocolate = true)
   assertThat(value.isHappy()).isTrue()
}


@Test
fun `check what makes SomeReallyComplicatedDataClassWith1000fields really sad`() {
   val value = dto.copy(isWeekend = true, isRaining = true)
   assertThat(value.isReallySad()).isTrue()
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
```diff
@@ -212,4 +212,7 @@ data class Licence(
 
   @Schema(description = "The full name of the person who created licence or variation", example = "Gordon Sumner")
   val createdByFullName: String? = null,
+
+  @Schema(description = "Is this licence in PSS period?", example = "(LED < TODAY <= TUSED)")
```
andrewrlee, line 216: example should be `true` or `false`, but you could add that rule in the description?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/TransformFunctions.kt
nishanthmahasamudram: moved isInPssPeriod to EntityLicence.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
nishanthmahasamudram: added all cases for unit testing isInPSSPeriod

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
```diff
@@ -163,4 +163,14 @@ data class Licence(
       dateCreated = LocalDateTime.now(),
     )
   }
+
+  fun isInPssPeriod(licence: Licence): Boolean {
```
andrewrlee, line 167: now that you have moved this function to licence, you don't need to pass through the licence to the function.

You have access to all this licences fields directly

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -142,6 +142,79 @@ class LicenceServiceTest {
     assertThat(licence.nomsId).isEqualTo(aLicenceEntity.nomsId)
     assertThat(licence.bookingId).isEqualTo(aLicenceEntity.bookingId)
     assertThat(licence.pnc).isEqualTo(aLicenceEntity.pnc)
+//    assertThat(licence.isInPssPeriod).isFalse()
```
andrewrlee, line 145: commented out code

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -142,6 +142,79 @@ class LicenceServiceTest {
...
+
+    verify(licenceRepository, times(1)).findById(1L)
+  }
+
+  @Test
+  fun `service transforms key fields of a licence should return isInPssPeriod false if LED is null`() {
```
andrewrlee, line 151: now you have the logic on licence you can add a `LicenceTest` that tests the functionality without having to be concerned about repositories or services

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
nishanthmahasamudram: removed.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
nishanthmahasamudram: Removed licence param.

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -698,6 +700,17 @@ class LicenceService(
     )
   }
 
+  @Transactional
+  fun removeAPConditions() {
+    val licencesInPSSPeriod = licenceRepository.getAllVariedLicencesInPSSPeriod()
+
+    licencesInPSSPeriod?.forEach { licence ->
+      val licenceWithAPConditions = licence.additionalConditions.filter { LicenceType.valueOf(it.conditionType!!) == LicenceType.AP }
```
andrewrlee, line 708: This will trigger an n+1 queries as it's iterating over each licence and retrieving the conditions lazily behind the scenes.

It would be good to add an integration test for this and we will be able to see this in the logs and check that improvements to the fetch strategy in the repositiory is working

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -698,6 +700,17 @@ class LicenceService(
     )
   }
 
+  @Transactional
+  fun removeAPConditions() {
```
andrewrlee, line 704: Can we move this into it's own service in the `jobs` package like `LicenceActivationService`?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -698,6 +700,17 @@ class LicenceService(
...
+    val licencesInPSSPeriod = licenceRepository.getAllVariedLicencesInPSSPeriod()
+
+    licencesInPSSPeriod?.forEach { licence ->
+      val licenceWithAPConditions = licence.additionalConditions.filter { LicenceType.valueOf(it.conditionType!!) == LicenceType.AP }
+
+      licenceWithAPConditions?.forEach { this.licenceConditionService.deleteAdditionalCondition(it.licence.id, it.id) }
```
andrewrlee, line 710: Think this will need to be some form of bulk update - otherwise you're doing an additional `n x a` licence loads and deletes where `n` is the number of licences and `a` is the number of additional conditions

andrewrlee: you could maybe add an overload of `deleteAdditionalCondition`:

```kt
 // old code calls this
  fun deleteAdditionalCondition(licenceId: Long, conditionId: Long) {
    val licenceEntity = licenceRepository
      .findById(licenceId)
      .orElseThrow { EntityNotFoundException("$licenceId") }
    this.deleteAdditionalCondition(licence, listOf(condtionId))
}

//your job calls this directly
 fun deleteAdditionalCondition(licenceId: Licence, conditionIds: List<Long>) {
    // ...modify existing code to filter out many conditions at once ...
}

```

Does that make sense? If not we can have a chat

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
```diff
@@ -1273,4 +1273,33 @@ class LicenceController(
   ) {
     updateSentenceDateService.updateSentenceDates(licenceId, request)
   }
+
+  @PostMapping(value = ["/run-remove-ap-conditions-job"])
```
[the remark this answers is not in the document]
nishanthmahasamudram: Created a new JobController and moved it.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
nishanthmahasamudram: created and moved in to RemoveApConditionsInPssPeriodService

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
nishanthmahasamudram: Yes, It makes sense.
I updated the function to n+1 where we will be calling deleteAdditionalConditions once per licence.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
nishanthmahasamudram: Added Integration test.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceConditionService.kt
```diff
@@ -255,6 +255,26 @@ class LicenceConditionService(
     }
   }
 
+  @Transactional
+  fun deleteAdditionalConditions(licenceId: Long, conditionIds: List<Long>) {
+    val licenceEntity = licenceRepository
+      .findById(licenceId)
+      .orElseThrow { EntityNotFoundException("$licenceId") }
```
andrewrlee, line 262: you already have the licence so don't need to re-load it here. 

Would suggest, moving the deleteAdditionalCondition function over to delegate to this function
```kt
fun deleteAdditionalCondition(licenceId: Long, conditionId: Long) 
  val licenceEntity = licenceRepository
      .findById(licenceId)
      .orElseThrow { EntityNotFoundException("$licenceId") }
  deleteAdditionalConditions(licenceEntity, listOf(conditionId))
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/RemoveApConditionsInPssPeriodService.kt
```diff
@@ -0,0 +1,31 @@
...
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @Transactional
+  fun removeAPConditions() {
+    log.info("Job to removeApConditions in PSS period started")
+    val licencesInPSSPeriod = licenceRepository.getAllVariedLicencesInPSSPeriod()
+
+    licencesInPSSPeriod?.forEach { licence ->
```
andrewrlee, line 25: this shouldn't be nullable so no need for the `?.` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
```diff
@@ -36,6 +36,22 @@ interface LicenceRepository : JpaRepository<Licence, Long>, JpaSpecificationExec
...
+                AND l.typeCode IN (uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceType.AP_PSS))
+        AND l.statusCode IN (
+            uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.VARIATION_IN_PROGRESS,
+            uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.VARIATION_SUBMITTED,
+            uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.VARIATION_REJECTED,
+            uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.VARIATION_APPROVED
+        )
+    """,
+  )
+  fun getAllVariedLicencesInPSSPeriod(): List<Licence>
```
andrewrlee, line 54: I think this is ok with out optimising - it does do multiple n*1 issues but it needs them for the delete and I suspect the number of licences that this will affect to be quite small. 

I think we'll need to observe this in preprod/prod

nishanthmahasamudram: yes, we can observe the performance in preprod/prod and then we can refactor if needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -764,7 +765,16 @@ class LicenceService(
     standardConditionRepository.saveAll(standardConditions)
     bespokeConditionRepository.saveAll(bespokeConditions)
 
-    val additionalConditions = licence.additionalConditions.map {
+    val licenceConditions: List<AdditionalCondition> =
+      if (newStatus == VARIATION_IN_PROGRESS) {
+        licence.additionalConditions.filter {
+          !(LicenceType.valueOf(it.conditionType!!) == LicenceType.AP && licence.typeCode == LicenceType.AP_PSS && licence.isInPssPeriod())
```
andrewrlee, line 771: this would be easier to reason about if we moved the ` licence.typeCode == LicenceType.AP_PSS && licence.isInPssPeriod()` into the surrounding if block? 

those are specific to the licence rather than each condition

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/RemoveApConditionsInPssPeriodService.kt
nishanthmahasamudram: removed.

andrewrlee (approved):
LGTM

PR #248, opened 2023-06-20
CVSL-1127 Adding condition formatter to the backend

andrewrlee:
This also fixes a couple of bugs:

* Incorrect casing of some inputs due to typo - captilise vs capitalised

* Incorrectly capitalise the joining word: or vs Or

This still needs to be hooked up, may dual run for a bit to identify any issues but then hope to remove the code from the frontend and move to stop exposing formatting rules

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionPss.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionPss.kt
@@ -8,6 +8,6 @@ data class AdditionalConditionPss(
   override val requiresInput: Boolean,
   override val categoryShort: String? = null,
   val pssDates: Boolean? = null,
-  override val inputs: List<PssInput>? = null,
+  override val inputs: List<Input>? = null,
   override val type: String? = null,
-) : IAdditionalCondition<PssInput>
+) : IAdditionalCondition
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/Conditional.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/Conditional.kt
@@ -2,4 +2,6 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy
 
 data class Conditional(
   val inputs: List<ConditionalInput>,
-)
+) : HasInputs {
+  override fun getConditionInputs() = inputs.map { it.toInput() }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/Input.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/Input.kt
@@ -1,17 +1,82 @@
 package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy
 
+import com.fasterxml.jackson.annotation.JsonValue
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.AdditionalConditionData
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy.InputType.ADDRESS
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy.InputType.TEXT
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.Case
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.adjustCase
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.formatAddress
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.formatUsing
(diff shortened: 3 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionPss.kt
```diff
@@ -8,6 +8,6 @@ data class AdditionalConditionPss(
   override val requiresInput: Boolean,
   override val categoryShort: String? = null,
   val pssDates: Boolean? = null,
-  override val inputs: List<PssInput>? = null,
```
andrewrlee, line 11: Remove the generic nature of these lists of inputs and removed the concept of PSSInputs.

This merely removed the ability to have the same level of rich formatting as Input which is a bit of artificial restriction

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/Conditional.kt
```diff
@@ -2,4 +2,6 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy
 
 data class Conditional(
   val inputs: List<ConditionalInput>,
-)
+) : HasInputs {
+  override fun getConditionInputs() = inputs.map { it.toInput() }
```
andrewrlee, line 6: Support a level of polymorphism to support using same mechanism to format conditional inputs as normal inputs

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/Input.kt
```diff
@@ -1,17 +1,82 @@
...
+  else -> formatMultipleValues(data)
+}
+
+private fun FormattingRule.formatSingleValue(item: String) = item
+  .adjustCase(this.case)
+  .let { this.applyFormattingRules(it) }
+
+private fun FormattingRule.formatMultipleValues(data: List<AdditionalConditionData>) = data
+  .mapNotNull { it.dataValue }
+  .map { it.adjustCase(this.case) }
```
andrewrlee, line 66: this is slightly different logic to the frontend - this would previously create the joined string of items and then apply casing rules which doesn't make sense for the joining word:

before: 
`[an, bn ,cn]` => `an, bn and cn` => `An, Bn, And Cn`
 
after:
`[an, bn ,cn]` => `an, bn and cn` => `An, Bn, and Cn`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies/PolicyV1.kt
```diff
@@ -148,18 +156,18 @@ val POLICY_V1_0 = LicencePolicy(
         inputs = listOf(
           Input(
             addAnother = AddAnother("Add another person"),
-            case = "capitalise",
+            case = CAPITALISED,
```
andrewrlee, line 159: Behaviour will change here - `capitalise` wasn't a valid case type so would silently ignore

Now this will correctly title case the Name.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ConditionFormatterTest.kt
```diff
@@ -0,0 +1,1043 @@
...
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy.InputType.TEXT
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy.InputType.TIME_PICKER
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy.Option
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.Case.CAPITALISED
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.Case.LOWER
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceType
+
+class ConditionFormatterTest {
+  private val conditionFormatter = ConditionFormatter()
```
andrewrlee, line 25: This test was ported over from the frontend and extended slightly to cover bugs

src/test/resources/test_data/policy_conditions/policyV2_1.json
```diff
@@ -949,7 +949,7 @@
                       "type": "text",
                       "label": "Enter the other frequency",
                       "name": "alternativeReportingFrequency",
-                      "case": "capitalise",
+                      "case": null,
```
andrewrlee, line 952: we decided to not fix this and keep the existing behaviour, as it didn't make sense to title case user provided durations

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/Input.kt
andrewrlee: got this confirmed by the designers

gurps-singh (approved):

PR #250, opened 2023-06-20
FB Automate Building API Extension

gurps-singh:
This PR takes the previous work around API automation one step further and creates a one stop script to build front and back end containers and run the back end from the API. There is still one step of manual operation to run the frontend from the root of the UI repo.

### DIFF
--- a/run-local.sh
+++ b/run-local.sh
@@ -39,7 +39,7 @@ docker compose down --remove-orphans
 #Prune existing containers
 #Comment in if you wish to perform a fresh install of all containers where all containers are removed and deleted
 #You will be prompted to continue with the deletion in the terminal
-docker system prune --all
+#docker system prune --all
 
 echo "Pulling back end containers ..."
 docker compose pull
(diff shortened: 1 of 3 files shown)


run-local.sh
```diff
@@ -32,9 +32,32 @@ export HMPPS_AUTH_URL=https://sign-in-dev.hmpps.service.justice.gov.uk/auth
...
+echo "Front end setup ..."
+echo "Bringing down current containers ..."
+docker compose down
+echo "Installing npm ..."
+npm install
+echo "Pulling front end containers ..."
+docker compose -f docker-compose-dev.yml pull
+docker compose -f docker-compose-dev.yml up -d
+echo "Waiting for front end containers to be ready ..."
+until [ "`docker inspect -f {{.State.Health.Status}} gotenberg`" == "healthy" ]; do
```
andrewrlee, line 46: rather than having the details about frontend dependencies in this project, could we create a run-local.sh script in the frontend? 

and maybe have a `run-full.sh` in this project that calls both run-local.sh scripts?

andrewrlee (approved):
LGTM

PR #251, opened 2023-06-22
CVSL-1127 tidying up updating additional conditions code

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceConditionService.kt
```diff
@@ -117,47 +117,27 @@ class LicenceConditionService(
...
+    val removedConditions = existingConditions.getRemovedConditions(submittedConditions, request)
 
-    submittedAdditionalConditions.forEach { newAdditionalCondition ->
-      updatedAdditionalConditions.filter {
-        it.conditionCode == newAdditionalCondition.conditionCode
-      }.forEach {
-        it.conditionCategory = newAdditionalCondition.conditionCategory
-        it.conditionText = newAdditionalCondition.conditionText
-        it.conditionSequence = newAdditionalCondition.conditionSequence
-        it.conditionType = newAdditionalCondition.conditionType
```
andrewrlee, line 147: not sure why this doesn't update the expanded text too!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceConditionService.kt
```diff
@@ -117,47 +117,27 @@ class LicenceConditionService(
...
-    val submittedAdditionalConditions =
+    val submittedConditions =
       request.additionalConditions.transformToEntityAdditional(licenceEntity, request.conditionType)
 
-    val newAdditionalConditions =
-      submittedAdditionalConditions.filter { licenceEntity.additionalConditions.none { submittedCondition -> submittedCondition.conditionCode == it.conditionCode } }
-        .map { newAdditionalCondition ->
-          newAdditionalCondition.copy(
-            expandedConditionText = newAdditionalCondition.conditionText,
-            licence = licenceEntity,
```
andrewrlee, line 128: don't need to set the licence as that's already done in the transform

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceConditionService.kt
```diff
@@ -117,47 +117,27 @@ class LicenceConditionService(
...
-          )
-        }
+    val existingConditions = licenceEntity.additionalConditions
 
-    val removedAdditionalConditionsList =
-      licenceEntity.additionalConditions.filter {
-        it.conditionType == request.conditionType && submittedAdditionalConditions.none { submittedCondition -> submittedCondition.conditionCode == it.conditionCode }
```
andrewrlee, line 134: not sure why this checks on conditionType whereas the other ones don't. 
I think we have unique codes across AP and PSS so this probably isn't necessary (I've kept it in)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceConditionService.kt
gurps-singh: I think this updating for expanded text is done in the `updateAdditionalConditionData` function in the same service. No idea why?

gurps-singh (approved):
Very useful for the auditing ticket I am working on so thanks!

PR #256, opened 2023-06-26
CVSL-1127 Only compare formatting when additional data is provided

andrewrlee:
Also split out toEntity and toModel transformers


justin-gardiner (approved):
lgtm

PR #257, opened 2023-06-26
CVSL-1127 Move to lighter model for updating additional conditions

andrewrlee:
This removes these fields:
```
 val id: Long? = -1,
 val text: String? = null,
 val expandedText: String? = null,
 val data: List<AdditionalConditionData> = emptyList(),
 val uploadSummary: List<AdditionalConditionUploadSummary> = emptyList(),
```

 this should be backwards compatible as jackson will ignore unknown fields (and they weren't being used anyway)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionRequest.kt
@@ -1,0 +1,19 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Describes an additional condition to create/update")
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionRequest.kt
```diff
@@ -0,0 +1,19 @@
...
+
+  @Schema(description = "The category of the additional condition", example = "Freedom of movement")
+  val category: String? = null,
+
+  @Schema(description = "Sequence of this additional condition within the additional conditions", example = "1")
+  val sequence: Int? = null,
+
+  @Schema(description = "The textual value for this additional condition", example = "You must not enter the location [DESCRIPTION]")
+  val text: String? = null,
+
```
andrewrlee, line 18: I think category and text could be derived from the code alone

I think index could possibly be derived from position in array as this overwrites all existing conditions

justin-gardiner (approved):
lgtm

PR #262, opened 2023-06-29
adding crn and com name to sendVariationForApprovalEmail to match the…

nishanthmahasamudram:
… new template


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -373,6 +373,8 @@ class LicenceService(
           licenceId.toString(),
           updatedLicence.forename!!,
           updatedLicence.surname!!,
+          licenceEntity.crn!!,
+          licenceEntity.submittedBy?.username!!,
```
andrewrlee, line 377: think `licenceEntity` would have the previous details on? we only update a copy of the licence with the submitter.

 think `username` should be fine?

nishanthmahasamudram: Even the submitter is getting updated:
val username = SecurityContextHolder.getContext().authentication.name
    val submitter = communityOffenderManagerRepository.findByUsernameIgnoreCase(username)

nishanthmahasamudram: updated.

andrewrlee (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -401,6 +401,8 @@ class LicenceService(
           licenceId.toString(),
           updatedLicence.forename!!,
           updatedLicence.surname!!,
+          licenceEntity.crn!!,
```
andrewrlee, line 404: no big deal but could have used `updatedLicence `

nishanthmahasamudram: updated

andrewrlee (approved):
LGTM

PR #264, opened 2023-06-29
Adding endpoint to update offender name and DOB

SebNegahbanMoJ:
Paired with UI PR [here](https://github.com/ministryofjustice/create-and-vary-a-licence/pull/575)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
@@ -134,13 +134,13 @@ class OffenderController(private val offenderService: OffenderService, private v
         responseCode = "403",
         description = "Forbidden, requires an appropriate role",
         content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
-      )
-    ]
+      ),
+    ],
   )
   fun updateOffenderDetails(
     @PathVariable nomsId: String,
-    @RequestBody body: UpdateOffenderDetailsRequest
-  ){
+    @RequestBody body: UpdateOffenderDetailsRequest,
+  ) {
     this.offenderService.updateOffenderDetails(nomsId, body)
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
@@ -140,7 +140,8 @@ class OffenderController(private val offenderService: OffenderService, private v
   )
   fun updateOffenderDetails(
     @PathVariable nomsId: String,
-    @Valid @RequestBody body: UpdateOffenderDetailsRequest,
+    @Valid @RequestBody
+    body: UpdateOffenderDetailsRequest,
   ) {
     this.offenderService.updateOffenderDetails(nomsId, body)
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
@@ -7,6 +7,7 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.responses.ApiResponses
 import io.swagger.v3.oas.annotations.security.SecurityRequirement
 import io.swagger.v3.oas.annotations.tags.Tag
+import jakarta.validation.Valid
 import org.springframework.http.MediaType
 import org.springframework.security.access.prepost.PreAuthorize
 import org.springframework.web.bind.annotation.PathVariable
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
```diff
@@ -105,4 +106,41 @@ class OffenderController(private val offenderService: OffenderService, private v
   ) {
     this.offenderService.updateProbationTeam(crn, body)
   }
+
+  @PutMapping(value = ["nomisid/{nomsId}/update-offender-details"])
```
andrewrlee, line 110: `nomsId` to match the other example? 

it should be `prisonNumber` but that's a whole 'nother kettle of fish

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
```diff
@@ -105,4 +106,41 @@ class OffenderController(private val offenderService: OffenderService, private v
...
+      ApiResponse(
+        responseCode = "403",
+        description = "Forbidden, requires an appropriate role",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  fun updateOffenderDetails(
+    @PathVariable nomsId: String,
+    @RequestBody body: UpdateOffenderDetailsRequest,
```
andrewrlee, line 142: Missing `@ Valid` here I think.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/OffenderService.kt
```diff
@@ -97,4 +98,51 @@ OffenderService(
       }
     }
   }
+
+  @Transactional
+  fun updateOffenderDetails(nomsId: String, request: UpdateOffenderDetailsRequest) {
+    val licenceStatuses = listOf(
```
andrewrlee, line 104: maybe the variable name could reflect what this list represents? I think it might mean "not inactive" licences?

In the past I've been on projects where we would centralise common lists of statuses as static fields in the Enum, potentially extending the domain if suitable `LicenceStatus.ALL_ACTIVE` `PolicyStatus.ALL_ON_TRIAL`  etc.. not sure we've hit that limit yet

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/OffenderService.kt
```diff
@@ -97,4 +98,51 @@ OffenderService(
...
+      LicenceStatus.SUBMITTED,
+      LicenceStatus.APPROVED,
+      LicenceStatus.VARIATION_IN_PROGRESS,
+      LicenceStatus.VARIATION_SUBMITTED,
+      LicenceStatus.VARIATION_APPROVED,
+      LicenceStatus.VARIATION_REJECTED,
+      LicenceStatus.ACTIVE,
+    )
+    val existingLicences = this.licenceRepository.findAllByNomsIdAndStatusCodeIn(nomsId, licenceStatuses)
+    var offenderDetailsChanged = false
```
andrewrlee, line 115: Mutating state in a `map` is a bit of a code smell... generally would try avoiding it in higher order functions.  

One way around this is just splitting up the operation which would break down the operation into declarative steps and lead to an overall simpler approach. 
```
  val licencesToChange = existingLicences.filter ...
  val updatedLicences = licencesToChange.map ...
  if (updatedLicences.isNotEmpty) {
     ...
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
SebNegahbanMoJ: `nomisId` is used more frequently in the UI, `nomsId` in the API 😅 I was flip-flopping back and forth on which to do, and am happy to change it to `nomsid` if you think it'd be better

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/OffenderService.kt
SebNegahbanMoJ: Agreed that while technically less performant, it will be unnoticeable.
I can't think of a scenario when someone would have more than two licences at once (now that `inactive` licences are excluded), so iterating over an array of 0-2 objects twice rather than once shouldn't cause the downfall of our app 😂

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderController.kt
andrewrlee: I think by example below I thought there was another endpoint that used it.. but I guess I meant the use in the pathVariable: "nomisid/{**nomsId**}"

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/OffenderControllerTest.kt
```diff
@@ -95,4 +97,23 @@ class OffenderControllerTest {
...
+
+    val request = put("/offender/nomisid/exampleNomisId/update-offender-details")
+      .accept(MediaType.APPLICATION_JSON)
+      .contentType(MediaType.APPLICATION_JSON)
+      .content(mapper.writeValueAsBytes(body))
+
+    mvc.perform(request).andExpect(status().isOk)
+
+    verify(offenderService, times(1)).updateOffenderDetails("exampleNomisId", body)
+  }
```
andrewrlee, line 118: should probably start testing that passing invalid payloads generates appropriate responses

andrewrlee (approved):
LGTM

PR #265, opened 2023-06-29
CVSL-1060: When updating a licence status to approved then deactivate…

justin-gardiner:
… any previous version.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -243,40 +247,40 @@ class LicenceService(
...
       else -> "Check - licence status not accounted for in auditing"
     }
 
     auditEventRepository.saveAndFlush(
       AuditEvent(
-        licenceId = licenceId,
-        username = request.username,
-        fullName = request.fullName,
-        eventType = getAuditEventType(request),
+        licenceId = licenceEntity.id,
```
SebNegahbanMoJ, line 272: If we only need the ID for the updated licence, could the passed param just be `updatedLicence.id`?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -321,6 +320,35 @@ class LicenceService(
     )
   }
 
+  private fun deactivatePreviousLicenceVersion(licence: EntityLicence, username: String, fullName: String?) {
```
andrewrlee, line 323: presumably this doesn't need to call itself recursively as this is effectively ensuring that only a single previous active licence will exist?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -803,4 +836,12 @@ class LicenceService(
 
     return newLicence
   }
+
+  private fun splitName(fullName: String?): Pair<String?, String?> {
+    val names = fullName?.split(" ")?.toMutableList()
+    val firstName = names?.firstOrNull()
+    names?.removeAt(0)
```
andrewrlee, line 843: super unnecessary comment (I know you're just moving this method!)

`drop(1).join...` would have got rid of the mutable list.

andrewrlee (approved):
LGTM

PR #266, opened 2023-07-06
CVSL-1148-CVL - Prevent Activation Job errors

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationService.kt
```diff
@@ -38,7 +39,11 @@ class LicenceActivationService(
 
   private fun determineLicenceEligibility(licences: List<Licence>): Pair<List<Licence>, List<Licence>> {
     val bookingIds = licences.map { it.bookingId!! }
-    val hdcStatuses = prisonApiClient.getHdcStatuses(bookingIds)
+    val hdcStatuses = if (bookingIds != null) {
```
SebNegahbanMoJ, line 42: I think this check can be taken higher
If `bookingIds` is null/empty, that means the `licences` param was empty, and if the `licences` param is empty there must not be any `potentialLicences` on line 31.
If that's the case, the entire job can be skipped, as there are no licences to activate.

My suggestion would be adding
```
if(potentialLicences.isEmpty()) { return }
```
on line 32. We may want to add a log stating that the job is ending because there were no licences, not sure.

andrewrlee: meh...  could pull it down into the client so less likely other people reusing these things - like the search endpoint to cause similar issues?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationService.kt
```diff
@@ -28,12 +27,18 @@ class LicenceActivationService(
   @Transactional
   fun licenceActivationJob() {
     val potentialLicences = licenceRepository.getApprovedLicencesOnOrPassedReleaseDate()
+    if (potentialLicences.isEmpty()) {
```
andrewrlee, line 30: can we have some test coverage for this?

andrewrlee (approved):
LGTM

PR #268, opened 2023-07-07
CVSL-1128 add caching for JWKS

gurps-singh:
This PR is to address the issue of repeated calls to the `hmpps-auth` endpoint unnecessarily for when the public key is rotated. This happens infrequently so we should cache. We have replicated what the `prisoner-offender-search` team have done around caching requests indefinitely.


src/main/resources/application-dev.yml
```diff
@@ -6,15 +6,8 @@
...
-spring:
-  security:
-    oauth2:
-      resourceserver:
-        jwt:
-          public-key-location: classpath:local-public-key.pub
-
-  jpa:
-    show-sql: true
+jpa:
```
petergphillips, line 9: this looks wrong - you've moved `jpa:` up a level to the root whereas it should still be under `spring:`

gurps-singh: Whoops, thanks for the spot Peter. Changed that.

andrewrlee (approved):
LGTM

PR #269, opened 2023-07-07
CVSL-1157-gradle-owasp-detekt-Vulnerability

nishanthmahasamudram:


build.gradle.kts
```diff
@@ -8,6 +8,9 @@ plugins {
 configurations {
   testImplementation { exclude(group = "org.junit.vintage") }
 }
+dependencyCheck {
```
nishanthmahasamudram, line 11: suppressing the false positive created by detekt

PR #276, opened 2023-07-24
CVSL-1229 Adding missing Valid annotations to enforce rules

andrewrlee:
Added a test to catch this happening in the future

Also adding the standard client tracking configuration code so we can determine who is calling our service for when we have future clients


louisenorris (approved):
LGTM

justin-gardiner (approved):
lgtm

PR #277, opened 2023-07-24
CVSL-1159: Add new endpoint to get recently approved licences.

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
```diff
@@ -444,6 +445,46 @@ class LicenceController(
...
+      ),
+      ApiResponse(
+        responseCode = "403",
+        description = "Forbidden, requires an appropriate role",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  fun getRecentlyApprovedLicences(
+    @RequestBody body: RecentlyApprovedLicencesRequest,
```
andrewrlee, line 481: this will need an Valid annotation or the test will fail.. 

I could have made the test more clever and tried to look for validation annotations on the payload itself but its quite a bit more complex so just enforced an RequestBody to also have Valid

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
```diff
@@ -36,6 +37,21 @@ interface LicenceRepository : JpaRepository<Licence, Long>, JpaSpecificationExec
...
+        WHERE l.prisonCode IN :prisonCodes 
+        AND l.statusCode IN (
+            uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.ACTIVE,
+            uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.APPROVED
+        )
+        AND (COALESCE(l.actualReleaseDate, l.conditionalReleaseDate) > :releasedAfterDate)
+        ORDER BY l.conditionalReleaseDate ASC
+    """,
+  )
+  fun getRecentlyApprovedLicences(prisonCodes: List<String>?, releasedAfterDate: LocalDate): List<Licence>
```
andrewrlee, line 54: is there a reason why this is nullable? 

what happens if null is passed?

justin-gardiner: I copied it from somewhere else, it doesn't make sense to be nullable here, fixed

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/LicenceController.kt
justin-gardiner: fixed

PR #286, opened 2023-07-28
CVSL-1094 enrich search response with licence data

gurps-singh:
This PR is to enrich the existing search response from the probation search with data available from our DB on licences. The PR filters out non relevant licences for search (i.e. inactive) and enriches the response with data on dates and statuses using the licence table.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/EnrichedProbationSearchResults.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation
+
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.ProbationSearchResult
+
+data class EnrichedProbationSearchResults(
+  val results: List<ProbationSearchResult>,
+  val inPrisonCount: Int,
+  val onProbationCount: Int,
+)
```
andrewrlee, line 9: is this being returned to the client? If so I think it should live in model and have swagger annotations?

gurps-singh: Good spot - I also forgot to update the swagger doc on the endpoint so will also do that.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/ProbationSearchResult.kt
```diff
@@ -1,8 +1,16 @@
...
+
 data class ProbationSearchResult(
   val name: Name,
   val identifiers: Identifiers,
   val manager: Manager,
   val allocationDate: String,
+  val licenceType: LicenceType? = null,
+  val licenceStatus: LicenceStatus? = null,
+  val releaseDate: LocalDate? = null,
+  val isOnProbation: Boolean? = null,
```
andrewrlee, line 15: I'd keep this class "clean" and just match what is returned from the API rather than have the enriched fields on here as a mechanism to pass the data across on the DTO. 


Think it'd be better to do the transform in the mapForNull and pass the licence to `transformToModelProbationResult` as a parameter?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -96,6 +99,39 @@ class ComService(
...
+
+    val enrichedProbationSearchResult = entityProbationSearchResult.mapNotNull {
+      val licences =
+        licenceRepository.findAllByCrnAndStatusCodeIn(it.identifiers.crn, LicenceStatus.searchRelevantLicenceStatus())
+
+      // If an empty list has been returned, there are no relevant licences relating to search for the offender
+      if (licences.isEmpty()) {
+        null
+      } else {
+        val nonActiveLicenceStatuses = LicenceStatus.searchRelevantLicenceStatus().minusElement(LicenceStatus.ACTIVE)
```
andrewrlee, line 111: think [minusElement](https://github.com/JetBrains/kotlin/blob/7a7d392b3470b38d42f80c896b7270678d0f95c3/libraries/stdlib/common/src/generated/_Collections.kt#L3178) is an alias for minus?

so can presumably call it as the `-` operator?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceStatus.kt
```diff
@@ -30,5 +30,28 @@ enum class LicenceStatus {
...
+        SUBMITTED,
+        APPROVED,
+        VARIATION_IN_PROGRESS,
+        VARIATION_SUBMITTED,
+        VARIATION_APPROVED,
+        VARIATION_REJECTED,
+      )
+    }
+
+    fun searchOnProbationLicenceStatus(): List<LicenceStatus> {
```
andrewrlee, line 47: could maybe have a field here = 
```kt
val PROBATION_VALUES = setOf(...
```

and then a helper function on enum of `isWithProbation`?
```kt
fun isWithProbation() = PROBATION_VALUES.contains(this)
```

which would simplify the logic where you call it?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceStatus.kt
```diff
@@ -30,5 +30,28 @@ enum class LicenceStatus {
...
+      return listOf(
+        ACTIVE,
+        IN_PROGRESS,
+        SUBMITTED,
+        APPROVED,
+        VARIATION_IN_PROGRESS,
+        VARIATION_SUBMITTED,
+        VARIATION_APPROVED,
+        VARIATION_REJECTED,
+      )
```
andrewrlee, line 44: We could raised the layer of abstraction here rather than name based on how they're used. 

I think this is the same list that we've named "inflight licences" in other parts of the code base, maybe that would be a good name for them here?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/EnrichedProbationSearchResults.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/ProbationSearchResult.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceStatus.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceStatus.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/ComController.kt
```diff
@@ -125,8 +124,8 @@ class ComController(private val comService: ComService, private val notifyServic
     value = [
       ApiResponse(
         responseCode = "200",
-        description = "The query retrieved a set of results",
-        content = [Content(mediaType = "application/json", array = ArraySchema(schema = Schema(implementation = ProbationSearchResult::class)))],
+        description = "The query retrieved a set of enriched results",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ProbationSearchResult::class))],
```
andrewrlee, line 128: need to update schema?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/model/EnrichedProbationSearchResult.kt
```diff
@@ -0,0 +1,17 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.model
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
+import java.time.LocalDate
+
+data class EnrichedProbationSearchResult(
```
andrewrlee, line 7: sorry I meant under: `uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model`

Think it needs swagger annotations too?

gurps-singh: So this is now the model used for the transform to turn each of the entity search results into the model enriched search results. There is an existing model class in `uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model` called `ProbationSearchResult` which I have repurposed so it caters for a list of these model enriched search results with the two relevant counts. These have swagger annotation as well.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -96,6 +99,33 @@ class ComService(
...
+        null
+      } else {
+        val nonActiveLicenceStatuses = LicenceStatus.IN_FLIGHT_LICENCES - LicenceStatus.ACTIVE
+
+        val currentLicence =
+          if (licences.size > 1) licences.find { licence -> licence.statusCode in nonActiveLicenceStatuses } else licences.first()
+
+        val isOnProbation = currentLicence?.statusCode?.isOnProbation()
+
+        transformToModelEnrichedSearchResult(it, currentLicence, isOnProbation)
```
andrewrlee, line 118: would probably make this an extension method on `it`. 
(You don't need to pass  `isOnProbation` as a separate field as you can access this on licence and this is just transformational logic - copying from the source field on to the destination field)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -1,8 +1,9 @@
 package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
 
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.Licence
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.LicenceSummary
-import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.ProbationSearchResult
-import java.util.Base64
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.ProbationSearchResponseResult
+import java.util.*
```
andrewrlee, line 6: shouldn't really be doing wild card imports - there's an option in intelliJ to disable these.
I'm not sure how it got through ktlint 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
gurps-singh: Done

PR #292, opened 2023-08-11
CVSL-999 add additional fields for search response

gurps-singh:
This PR is to add some additional fields for the search response from the API to enable the results table to be displayed accurately in the frontend.


SebNegahbanMoJ (approved):
Looks reasonable to me!

PR #296, opened 2023-08-18
CVSL-1204: Add licence_version column to licence table. Version is in…

justin-gardiner:
…itially set to 1.0 for new licences. When editing an approved licence the minor version will be incremented, for example, 1.1, 1.2. When creating a new variation the major version will be incremented, for example, 2, 3.


justin-gardiner:
This should not be merged until we have a way of populating the version numbers of existing licences. This is being implemented by https://dsdmoj.atlassian.net/browse/CVSL-1205

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -908,4 +910,25 @@ class LicenceService(
     }
     return originalLicence
   }
+
+  private fun getNextLicenceVersion(currentVersion: String, status: LicenceStatus): String {
+    val (majorVersion, minorVersion) = getVersionParts(currentVersion)
+    return when (status) {
+      VARIATION_IN_PROGRESS -> (majorVersion + 1).toString()
```
andrewrlee, line 917: so we don't have a `.0` on variations.. presumably this is intentional?

andrewrlee (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -778,13 +778,15 @@ class LicenceService(
     val createdBy = this.communityOffenderManagerRepository.findByUsernameIgnoreCase(username)
       ?: throw IllegalStateException("Cannot find staff with username: $username")
 
-    val licenceCopy = licence.copyLicence(newStatus)
+    val newLicenceVersion =
+      licence.licenceVersion?.let { getNextLicenceVersion(it, newStatus) }
+
+    val licenceCopy = licence.copyLicence(newStatus, newLicenceVersion)
```
andrewrlee, line 784: this method is humongous now!..

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
justin-gardiner: yes that was what they wanted. I did think of keeping the .0 to make it simpler and could then just change how it's displayed in the frontend.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
justin-gardiner: yes, will have a look to see if there is anything we can extract out

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
SebNegahbanMoJ: Ah, the script that I've written to populate existing versions will still have `.0` on the end. I can change it to remove that if this is the approach we'd rather take!

justin-gardiner: ok, as discussed I will update this, it will make this code simpler and if required we can just strip the .0 off variations in the front end.

justin-gardiner: updated

SebNegahbanMoJ (approved):

PR #302, opened 2023-08-30
CVSL-1298 Adding logging to licence activation job to show offence histories

SebNegahbanMoJ:
Tested locally by running existing specs that called the affected method, and the output was as expected.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/IS91DeterminationService.kt
```diff
@@ -20,6 +22,8 @@ class IS91DeterminationService(
         resultCodes.contains(it.secondaryResultCode) ||
         it.offenceCode == offenceCode
     }
+    log.info("All offence histories: ${offenceHistories.toString()}")
```
andrewrlee, line 25: meh.. for future reference, you don't need to explicitly call `toString`, its called by default when trying to print objects

andrewrlee (approved):
LGTM

PR #305, opened 2023-09-06
CVSL-1320 deactivate in progress licence versions

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -924,4 +933,15 @@ class LicenceService(
     val parts = version.split(".")
     return Pair(parts[0].toInt(), parts[1].toInt())
   }
+
+  private fun inactivateInProgressLicenceVersions(licences: List<EntityLicence>) {
+    val licenceIds = licences.map { it.id }
+    val licencesToDeactivate =
+      licenceRepository.findAllByVersionOfIdInAndStatusCodeIn(licenceIds, listOf(IN_PROGRESS, SUBMITTED))
```
andrewrlee, line 940: is it worth capturing the reason why we're inactivating them or is it going to be reasonably apparent?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -528,6 +534,9 @@ class LicenceService(
           ),
         )
       }
+      inactivateInProgressLicenceVersions(
```
andrewrlee, line 537: some recursion here?
`inactivateInProgressLicenceVersions -> inactivateLicences  -> inactivateInProgressLicenceVersions`

justin-gardiner: yes there is recursion but it stops quickly as in progress versions don't have any versions linked to them. It works ok when I tested locally. I did add an optional param to stop the recursion when I was looking at ways around this but took it out, I'll add it back

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
justin-gardiner: yes, again I did have this before when I was calling this for each licence. I did an optimisation to process a list of all licences at once and removed the reason string for some reason. I'll have a look at adding it back

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
andrewrlee: if you think its fine, then thats' good enough!

justin-gardiner: updated to avoid the recursion to be safe

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
justin-gardiner: worth doing as it will make support easier, so added a reason param

andrewrlee (approved):
LGTM

PR #307, opened 2023-09-11
CVSL-1438: Update query to find unapproved licences to include licenc…

justin-gardiner:
…es with status IN_PROGRESS but don't have an approved audit event because they are versions on an existing approved licence.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
```diff
@@ -22,8 +22,10 @@ interface LicenceRepository : JpaRepository<Licence, Long>, JpaSpecificationExec
         JOIN l.submittedBy com
         WHERE (l.actualReleaseDate = CURRENT_DATE OR l.conditionalReleaseDate = CURRENT_DATE) 
         AND l.statusCode = 'SUBMITTED'
-        AND EXISTS 
-                (SELECT 1 FROM AuditEvent ae WHERE l.id = ae.licenceId AND ae.detail LIKE '%APPROVED%')
+        AND (
+            (EXISTS (SELECT 1 FROM AuditEvent ae WHERE l.id = ae.licenceId AND ae.detail LIKE '%APPROVED%')) 
+         OR (l.versionOfId IS NOT NULL) 
+        )
     """,
```
justin-gardiner, line 29: This query finds in progress licences that have been previously approved. When versioning is enabled in progress licence versions won't have an approved audit event as this will be linked to the previous licence version but they will have versionOfId populated.

justin-gardiner, line 29: This query finds in progress licences that were previously approved. Before versioning any in progress licence that was previously approved will have an approved audit event.  After versioning is enabled in progress version won't have an approved audit event but will have a versionOfId value.

SebNegahbanMoJ: Am I right in thinking this won't select licences that are `IN_PROGRESS` (even if they're edits)? Is that intentional?

justin-gardiner: ahh, I see my comment above mentioned IN_PROGRESS when this only looks at SUBMITTED licences :) Yes you are right that this only selects licences that have been previously approved and now have a status of SUBMITTED. it doesn't return any IN_PROGRESS licences that have been previously approved. It behaved like this before so I assume that was correct behaviour, but I'm now thinking it might not be.

justin-gardiner: I don't think this was right before, going to check with Laura what it should be doing

justin-gardiner: ok, discussed with Laura and it was intended to only send notifications for SUBMITTED licences is the correct behaviour.

SebNegahbanMoJ (approved):

PR #309, opened 2023-09-13
CVSL-1364 Adding the rest of the sentence dates to LicenceSummary

SebNegahbanMoJ:


andrewrlee (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -1898,6 +1898,12 @@ class LicenceServiceTest {
       probationTeamDescription = "Cardiff South Team A",
       conditionalReleaseDate = LocalDate.of(2021, 10, 22),
       actualReleaseDate = LocalDate.of(2021, 10, 22),
+      sentenceStartDate = LocalDate.of(2018, 10, 22),
+      sentenceEndDate = LocalDate.of(2021, 10, 22),
+      licenceStartDate = LocalDate.of(2021, 10, 22),
+      licenceExpiryDate = LocalDate.of(2021, 10, 22),
+      topupSupervisionStartDate = LocalDate.of(2021, 10, 22),
+      topupSupervisionExpiryDate = LocalDate.of(2021, 10, 22),
```
andrewrlee, line 1906: as a tip - its worth having unique values in test data..
It makes accidental typos pop out more

As an example: if you'd accidentally set `topupSupervisionStartDate` and `topupSupervisionEndDate` to the same value in your converter for instance

SebNegahbanMoJ: Good point!
I nabbed these from the `aLicenceEntity` above but I probably should've fixed both

PR #311, opened 2023-09-14
CVSL-1364 Removing filtering to APPROVED licences for licence deactivation endpoint

SebNegahbanMoJ:
The only call to this endpoint currently comes from the licenceExpiryService, which deactivates active licences, so the limitation of only finding `APPROVED` licences needs to be removed.


andrewrlee (approved):
LGTM, obviously not covered by a test!

SebNegahbanMoJ:
It is actually covered by a test, but the test didn't cover the fail-state, only that an `APPROVED` licence would be deactivated.
I can add some to cover the other states now, but there isn't a fail-state anymore, so the test now would just be checking that it passes through the mock-returned licences, which seemed a bit redundant given the simplicity of the code

PR #313, opened 2023-09-14
CVSL-999 Changes following QA feedback

gurps-singh:
This PR is to allow the retrieval of search results where the corresponding person does not currently have a licence. It includes logic to return a search result without licence details as well as associated changes to tests and logic in the `ComService`.

### DIFF
--- a/detekt-baseline.xml
+++ b/detekt-baseline.xml
@@ -7,6 +7,7 @@
     <ID>FunctionNaming:JPASpecificationDSL.kt$fun &lt;T, R : Any&gt; KProperty1&lt;T, R?&gt;.`in`(values: Collection&lt;R&gt;): Specification&lt;T&gt;</ID>
     <ID>LargeClass:LicenceController.kt$LicenceController</ID>
     <ID>LargeClass:LicenceService.kt$LicenceService</ID>
+    <ID>LongMethod:ComService.kt$ComService$fun searchForOffenderOnStaffCaseload(body: ProbationUserSearchRequest): ProbationSearchResult</ID>
     <ID>LongMethod:LicenceService.kt$LicenceService$@Transactional fun updateLicenceStatus(licenceId: Long, request: StatusUpdateRequest)</ID>
     <ID>LongMethod:LicenceService.kt$LicenceService$private fun copyLicenceAndConditions(licence: EntityLicence, newStatus: LicenceStatus): EntityLicence</ID>
     <ID>LongMethod:ToModelTransformers.kt$fun transform(licence: EntityLicence): ModelLicence</ID>
@@ -46,7 +47,6 @@
     <ID>UnusedParameter:LicenceController.kt$LicenceController$@PathVariable conditionType: String</ID>
     <ID>UnusedParameter:OpenApiConfiguration.kt$OpenApiConfiguration$buildProperties: BuildProperties</ID>
     <ID>UnusedParameter:WebClientConfiguration.kt$WebClientConfiguration$authorizedClientManager: OAuth2AuthorizedClientManager</ID>
-    <ID>UnusedPrivateProperty:ComService.kt$ComService$val prisonerSearchResult = this.prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(it.identifiers.noms))</ID>
     <ID>UseCheckOrError:CommunityApiClient.kt$CommunityApiClient$throw IllegalStateException("Unexpected null response from API")</ID>
     <ID>UseCheckOrError:LicenceService.kt$LicenceService$throw IllegalStateException("Can only update licence version when editing an existing version or creating a variation")</ID>
     <ID>UseCheckOrError:LicenceService.kt$LicenceService$throw IllegalStateException("Cannot find staff with username: $username")</ID>
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
@@ -14,7 +14,9 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.C
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.ProbationSearchApiClient
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.model.request.ProbationSearchSortByRequest
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceType
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.SearchDirection
+import java.time.LocalDate
 import java.time.LocalDateTime
 
 @Service
@@ -105,18 +107,68 @@ class ComService(
     val enrichedProbationSearchResults = entityProbationSearchResult.mapNotNull {
       val licences =
         licenceRepository.findAllByCrnAndStatusCodeIn(it.identifiers.crn, LicenceStatus.IN_FLIGHT_LICENCES)
-      val prisonerSearchResult = this.prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(it.identifiers.noms))
 
-      // If an empty list has been returned, there are no relevant licences relating to search for the offender
+      val comName =
+        if (it.manager.name?.forename.isNullOrBlank() || it.manager.name?.surname.isNullOrBlank()) {
+          null
+        } else {
+          "${it.manager.name?.forename} ${it.manager.name?.surname}".convertToTitleCase()
+        }
+
+      /**
+       *  If an empty list has been returned, we do not have a licence for the person currently but still need to return
+       *  a result for the PP as they may want to create a licence for them
+       */
       if (licences.isEmpty()) {
-        null
+        val isNomisNumberPopulated = !it.identifiers.noms.isNullOrBlank()
+
+        if (isNomisNumberPopulated) {
+          val prisonerSearchResults = this.prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(it.identifiers.noms))
+
+          if (prisonerSearchResults.isEmpty()) {
+            null
+          } else {
+            val prisonerSearchResult = prisonerSearchResults.first()
+
+            val licenceType = if (prisonerSearchResult.licenceExpiryDate.isNullOrBlank()) {
+              LicenceType.PSS
+            } else if (prisonerSearchResult.topUpSupervisionExpiryDate.isNullOrBlank() || LocalDate.parse(
+                prisonerSearchResult.topUpSupervisionExpiryDate,
+              ) <= LocalDate.parse(prisonerSearchResult.licenceExpiryDate)
+            ) {
+              LicenceType.AP
+            } else {
+              LicenceType.AP_PSS
+            }
+
+            // if both dates are null from the prisoner, we do not want to show the result as a licence cannot be created without them
+            if (prisonerSearchResult.confirmedReleaseDate.isNullOrBlank() && prisonerSearchResult.releaseDate.isNullOrBlank()) {
+              null
+            } else {
+              val releaseDate =
+                if (prisonerSearchResult.confirmedReleaseDate.isNullOrBlank()) {
+                  LocalDate.parse(prisonerSearchResult.releaseDate)
+                } else {
+                  LocalDate.parse(prisonerSearchResult.confirmedReleaseDate)
+                }
+              it.transformToModelFoundProbationRecordWithoutLicence(
+                comName,
+                releaseDate,
+                licenceType,
+                LicenceStatus.NOT_STARTED,
+              )
+            }
+          }
+        } else {
+          null
+        }
       } else {
         val nonActiveLicenceStatuses = LicenceStatus.IN_FLIGHT_LICENCES - LicenceStatus.ACTIVE
 
         val currentLicence =
           if (licences.size > 1) licences.find { licence -> licence.statusCode in nonActiveLicenceStatuses } else licences.first()
 
-        it.transformToModelFoundProbationRecord(currentLicence)
+        it.transformToModelFoundProbationRecord(currentLicence, comName)
       }
     }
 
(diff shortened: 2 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -105,18 +107,68 @@ class ComService(
...
+
+        if (isNomisNumberPopulated) {
+          val prisonerSearchResults = this.prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(it.identifiers.noms))
+
+          if (prisonerSearchResults.isEmpty()) {
+            null
+          } else {
+            val prisonerSearchResult = prisonerSearchResults.first()
+
+            val licenceType = if (prisonerSearchResult.licenceExpiryDate.isNullOrBlank()) {
```
gurps-singh, line 133: This logic is mirrored from the logic to ascertain a licence in the frontend - can someone just double check my logic here - current logic can be found [here](https://github.com/ministryofjustice/create-and-vary-a-licence/blob/f0b53be71cef78d6aaf66262f2e08dba69aeb145/server/services/licenceService.ts#L596).

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -105,11 +107,53 @@ class ComService(
...
+        val isNomisNumberPopulated = !it.identifiers.noms.isNullOrBlank()
+
+        if (isNomisNumberPopulated) {
+          val prisonerSearchResults = this.prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(it.identifiers.noms))
+
+          if (prisonerSearchResults.isEmpty()) {
+            null
+          } else {
+            val prisonerSearchResult = prisonerSearchResults.first()
+
```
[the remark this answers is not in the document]
andrewrlee: probably worth extracting out into a separate function and testing in isolation?

detekt-baseline.xml
```diff
@@ -7,6 +7,7 @@
     <ID>FunctionNaming:JPASpecificationDSL.kt$fun &lt;T, R : Any&gt; KProperty1&lt;T, R?&gt;.`in`(values: Collection&lt;R&gt;): Specification&lt;T&gt;</ID>
     <ID>LargeClass:LicenceController.kt$LicenceController</ID>
     <ID>LargeClass:LicenceService.kt$LicenceService</ID>
+    <ID>LongMethod:ComService.kt$ComService$fun searchForOffenderOnStaffCaseload(body: ProbationUserSearchRequest): ProbationSearchResult</ID>
```
andrewrlee, line 10: This is providing some good advice, worth thinking of ways to address it rather than add an exclusion

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -243,7 +246,7 @@ fun ProbationSearchResponseResult.transformToModelFoundProbationRecord(licence:
     name = "${name.forename} ${name.surname}".convertToTitleCase(),
     crn = licence?.crn,
     nomisId = licence?.nomsId,
-    comName = "${manager.name?.forename} ${manager.name?.surname}".convertToTitleCase(),
+    comName = if (manager.name?.forename.isNullOrBlank() || manager.name?.surname.isNullOrBlank()) null else "${manager.name?.forename} ${manager.name?.surname}".convertToTitleCase(),
```
andrewrlee, line 249: It looks like `manager.name` is the nullable thing?

If so, it's more idiomatic to:
`manage.name?.let { "${it.forename} ${it.surname}".convertToTitleCase()" }`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -253,3 +256,23 @@ fun ProbationSearchResponseResult.transformToModelFoundProbationRecord(licence:
     isOnProbation = licence?.statusCode?.isOnProbation(),
   )
 }
+
+fun ProbationSearchResponseResult.transformToModelFoundProbationRecordWithoutLicence(
```
andrewrlee, line 260: `transformToModelFoundProbationRecordWithoutLicence` -> `transformToUnstartedRecord`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/model/request/PrisonerSearchByPrisonerNumbersRequest.kt
```diff
@@ -1,5 +1,5 @@
 package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.model.request
 
 data class PrisonerSearchByPrisonerNumbersRequest(
-  val prisonerNumbers: List<String>,
+  val prisonerNumbers: List<String?>,
```
andrewrlee, line 4: mmmh... it might be better to filter out nulls if possible, rather than send them downstream

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -105,11 +107,53 @@ class ComService(
...
+      /**
+       *  If an empty list has been returned, we do not have a licence for the person currently but still need to return
+       *  a result for the PP as they may want to create a licence for them
+       */
       if (licences.isEmpty()) {
-        null
+        val isNomisNumberPopulated = !it.identifiers.noms.isNullOrBlank()
+
+        if (isNomisNumberPopulated) {
+          val prisonerSearchResults = this.prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(it.identifiers.noms))
```
andrewrlee, line 119: we probably don't want to do a separate call for each search result.. would be better to batch them 🤔

andrewrlee: So you've currently got: 
```
val records = ...

for r in records {
    val licences = ...
    if (licence.isEmpty) {
         createNewRecord(r)
    } else {
         createExistingRecord(r)
    }
}
```

andrewrlee: I'd possibly be tempted to 

```
val recordsWithLicences: List<Pair<Record, Licence?>> = records.map { it, getLicence(it) }
```
where getLicence, gets the relevant licence if one exists

Then:
```
val prisonersForUnstartedLicences: List<Prisoner search results> = ...findPrisoners(recordsWithLicences.filter { it.second == null }) 
``` 

Then:
```
recordsWithLicences.map { (record, licence) ->
   it (licence == null)  
      createUnstartedRecord(record, prisonersForUnstartedLicences.find { it.prisonNumber = record.prisonNumber})
  else
      createUnstartedRecord(record, prisonersForUnstartedLicences.find { it.prisonNumber = record.prisonNumber})
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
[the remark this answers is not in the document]
gurps-singh: Done

detekt-baseline.xml
gurps-singh: Based on splitting out logic into local extension functions, no longer an issue

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
gurps-singh: Magic, thanks Andy.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/PrisonerSearchPrisoner.kt
```diff
@@ -4,4 +4,8 @@ data class PrisonerSearchPrisoner(
   val prisonerNumber: String,
   val bookingId: String,
   val status: String? = null,
+  val licenceExpiryDate: String? = null,
```
andrewrlee, line 7: You can push the responsibility of deserialization into jackson by making these `LocalDate`

This principle of pushing type checking to the boundaries of an application is really nice and you can see the effect of not doing this in the frontend - date parsing everywhere!

Might need to add `@JsonFormat(pattern = "yyyy-MM-dd")`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -129,4 +134,77 @@ class ComService(
...
+      LicenceType.PSS
+    } else if (this.topUpSupervisionExpiryDate.isNullOrBlank() || LocalDate.parse(
+        this.topUpSupervisionExpiryDate,
+      ) <= LocalDate.parse(this.licenceExpiryDate)
+    ) {
+      LicenceType.AP
+    } else {
+      LicenceType.AP_PSS
+    }
+  }
```
andrewrlee, line 149: this `if return / else if return / else return` is often much nicer handled by pattern matching.

With date parsing pushed to the boundary, it could look something like:

```
  private fun PrisonerSearchPrisoner.getLicenceType() = when {
    licenceExpiryDate == null -> PSS
    topUpSupervisionExpiryDate == null || topUpSupervisionExpiryDate <= licenceExpiryDate -> AP
    else -> AP_PSS
  }
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -129,4 +134,77 @@ class ComService(
...
+
+  private fun PrisonerSearchPrisoner.getReleaseDate(): LocalDate? {
+    return if (this.confirmedReleaseDate.isNullOrBlank()) {
+      LocalDate.parse(this.releaseDate)
+    } else {
+      LocalDate.parse(this.confirmedReleaseDate)
+    }
+  }
```
andrewrlee, line 157: pushing to the boundary and using null handling: 
```
  private fun PrisonerSearchPrisoner.getReleaseDate() = confirmedReleaseDate ?: releaseDate
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -102,21 +108,20 @@ class ComService(
       probationSearchApiSortBy,
     )
```
andrewrlee, line 110: I'd extract the building up of `probationSearchApiSortBy` into an extension function.

This would help normalise the level of abstraction in this function so it's roughly constant across.

andrewrlee: Also maybe rename `entityProbationSearchResult` -> `teamCaseloadResults` ?

(ProbationSearchResult is naming based on type rather role and entity is generally used to refer to DB entities )

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -129,4 +134,77 @@ class ComService(
...
+        val releaseDate = prisoner.getReleaseDate()
+        result.transformToUnstartedRecord(
+          releaseDate,
+          licenceType,
+          LicenceStatus.NOT_STARTED,
+        )
+      }
+    }
+    return null
+  }
```
andrewrlee, line 205: might be nicer using pattern matching: 
```kt
  private fun ProbationSearchResponseResult.createUnStartedRecord(
    prisoner: PrisonerSearchPrisoner?,
  ) = when {
    // no match for prisoner in delius
    prisoner == null -> null

    // if both dates are null from the prisoner, we do not want to show the result as a licence cannot be created without them
    prisoner.confirmedReleaseDate == null && prisoner.releaseDate == null -> null

    // using match
    else -> this.transformToUnstartedRecord(
      releaseDate = prisoner.getReleaseDate(),
      licenceType = prisoner.getLicenceType(),
      LicenceStatus.NOT_STARTED,
    )
  }

```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/PrisonerSearchPrisoner.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -129,4 +134,77 @@ class ComService(
...
+        val prisonerSearchResult =
+          this.prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(result.identifiers.noms))
+        if (prisonerSearchResult.isEmpty()) {
+          null
+        } else {
+          prisonerSearchResult.first()
+        }
+      } else {
+        null
+      }
```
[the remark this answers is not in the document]
gurps-singh: Pattern matching wizardy - thanks Andy!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
gurps-singh: Done

PR #317, opened 2023-09-22
CVSL-1274 Adding public api to swagger

SebNegahbanMoJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/Tags.kt
```diff
@@ -1,4 +1,4 @@
-package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.resource
```
andrewrlee, line 1: This Tags and ControllerAnnotations could sit at the higher level as they would potentially be shared by both private and public apis

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/AuditControllerTest.kt
```diff
@@ -25,6 +25,7 @@ import org.springframework.test.web.servlet.setup.MockMvcBuilders
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.config.ControllerAdvice
```
andrewrlee, line 25: Should probably package these in the same way as their associated controllers

andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #319, opened 2023-09-25
CVSL-1380 Adding detekt to circle ci

andrewrlee:


.scripts/pre-commit/pre-commit
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+for hook in "$(dirname "$0")/pre-commit-scripts"/*; do
+   "$hook"
+   RESULT=$?
+
+   if [ $RESULT != 0 ]; then
```
justin-gardiner, line 7: should this be -ne like the exit status below?

gurps-singh (approved):

.scripts/pre-commit/pre-commit
justin-gardiner: actually just checked and both ways seem to work

.scripts/pre-commit/pre-commit-scripts/detekt.sh
```diff
@@ -0,0 +1,34 @@
...
+
+   echo "***********************************************"
+   echo "                 detekt passed                 "
+   echo "    No issues need fixing before committing    "
+   echo "***********************************************"
+
+   exit $EXIT_CODE
+fi
+
+rm "$OUTPUT"
```
justin-gardiner, line 34: It doesn't look like we can ever reach the rm as both branches of if exit

justin-gardiner: do you still want the exit $EXIT_CODE after the rm or doesn't it matter?

justin-gardiner: sorry just seen it is there!

justin-gardiner (approved):
lgtm

PR #324, opened 2023-09-26
CVSL-1165: Update the out of bounds premises condition to allow the u…

justin-gardiner:
…ser to enter just a name, just an address or a name and a address.


justin-gardiner:
This requires front end changes in https://github.com/ministryofjustice/create-and-vary-a-licence/pull/679 to work.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies/PolicyV2Dot1.kt
```diff
@@ -525,14 +525,52 @@ val POLICY_V2_1 = LicencePolicy(
...
+                  ),
+                ),
+              ),
+              Option(
+                value = "A name or type of premises and an address",
+                conditional = Conditional(
+                  inputs = listOf(
+                    ConditionalInput(
+                      label = "Name or type of premises",
+                      name = "nameOfPremises",
```
SebNegahbanMoJ, line 561: If you haven't already, you may need to test what happens if you enter a value either here or in `Name or type of premises` then go back via the `Change` link and switch to the other option.
If I recall, a similar change that I made caused problems when that behaviour happened, as both inputs shared the same `name`.
It may not be an issue for you here though 🤞

justin-gardiner: yeah I had to put some handling in the front end as I have two name and two address inputs on the page with the same name. So I added some js to disable the hidden inputs when the form is submitted. 

But I have now tested the back button and found another bug, due to the values being undefined, so will fix that.

andrewrlee (approved):
LGTM

PR #325, opened 2023-09-27
Revert "CVSL-1250 send the bank holiday object instead of list of dates"

gurps-singh:
Reverts ministryofjustice/create-and-vary-a-licence-api#323


andrewrlee (approved):
LGTM

PR #326, opened 2023-09-28
Adding representation

andrewrlee:


licence.json
```diff
@@ -0,0 +1,78 @@
+{
+  "id": 99999,
+  "type": "STD | HDC",
+  "subType": "AP | PSS | AP+PSS",
```
andrewrlee, line 4: Could do with better names for these?

licence.json
```diff
@@ -0,0 +1,79 @@
...
+  "approvedDate": "2023-09-27T16:59:04.886Z",
+  "createdByUsername": "X12333",
+  "createdDate": "2023-09-27T16:59:04.886Z",
+  "updatedByUsername": "X34433",
+  "updatedDate": "2023-09-27T16:59:04.886Z",
+  "isInPssPeriod": true,
+  "appointment": {
+    "contact": "Duty officer",
+    "time": "2023-09-27T16:59:04.886Z",
+    "address": "Manchester Probation Service, Unit 4, Smith Street, Stockport, SP1 3DN",
```
justin-gardiner, line 21: just wondering if it's worth having an address type? I know this is how we store addresses in cvl but at the moment this doesn't handle a "," in the address

gurps-singh: +1

licence.json
```diff
@@ -0,0 +1,79 @@
...
+  "type": "STD | HDC",
+  "subType": "AP | PSS | AP+PSS",
+  "policyVersion": "2.1",
+  "version": "1.4",
+  "statusCode": "ACTIVE",
+  "nomsId": "A9999AA",
+  "bookingId": 989898,
+  "crn": "X12444",
+  "approvedByUsername": "X33221",
+  "approvedDate": "2023-09-27T16:59:04.886Z",
```
gurps-singh, line 12: Could we not just provide the date (or is the timestamp necessary)

policy.json
```diff
@@ -0,0 +1,39 @@
...
+        {
+          "code": "9ce9d594-e346-4785-9642-c87e764bee37",
+          "text": "Be of good behaviour and not behave in a way which undermines the purpose of the licence period."
+        }
+      ],
+      "additional": [
+        {
+          "code": "5db26ab3-9b6f-4bee-b2aa-53aa3f3be7dd",
+          "category": "Residence at a specific place",
+          "categoryShort": null,
```
gurps-singh, line 32: Have included the `categoryShort` but still needs a discussion I think as to whether it is actually needed or could we just switch all places that use it to category?

licence.json
andrewrlee: Think it makes sense to have the timestamp as well - as these refer to events that happen at a specific time. 

But would could maybe call them `*DateTime`, e.g: `ApprovedDateTime`?

andrewrlee: We'll need to add something to the docs to explain that licences approved before X will not have an approved time.

licence.json
andrewrlee: Yes, agree but that's a slightly wider problem that maybe it would be better to resolve outside the scope of this work.

I think lets remove it for now and we can add it back in once we've resolved that issue

policy.json
andrewrlee: We need it in the frontend as its needed to vary the summary based on the context

We could possibly use the same logic as the frontend - return short if set else return long

I think exposing the value as show on the licence might be a sensible default

andrewrlee: We'll go with `category = categoryShort || category` logic for now

licence.json
andrewrlee: `licenceType`

PR #329, opened 2023-10-03
CVSL-1277 add representations for licence and policy

gurps-singh:
This PR is to add the public API representations of the licence and the policy.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/AdditionalConditionData.kt
```diff
@@ -0,0 +1,17 @@
...
+@Schema(description = "Describes each additional condition on the licence")
+data class AdditionalConditionData(
+
+  @Schema(description = "The additional condition's unique code", example = "abcd-1234-efgh")
+  val code: String,
+
+  @Schema(description = "The category to which the additional condition belongs", example = "Residence at a specific place")
+  val category: String,
+
+  @Schema(description = "The instances of an additional condition where multiple versions of the same condition exist")
```
gurps-singh, line 14: Couldn't think of a better description for `instances`

andrewrlee (approved):
LGTM

PR #331, opened 2023-10-05
CVSL-1277 Little bit of tidying up of representations

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/LicenceStatus.kt
```diff
@@ -0,0 +1,14 @@
...
+  SUBMITTED,
+  APPROVED,
+  ACTIVE,
+  REJECTED,
+  INACTIVE,
+  VARIATION_IN_PROGRESS,
+  VARIATION_SUBMITTED,
+  VARIATION_APPROVED,
+  ;
+}
```
andrewrlee, line 14: Rather than use the other set of statuses with some unimplemented statuses, just exposing what we use.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/Licence.kt
```diff
@@ -44,28 +43,28 @@ data class Licence(
   @Schema(
     description = "The date and time that the prison approved the licence, where licences approved before 01/04/2022 " +
       "will not have an approved time",
-    example = "24/08/2022 11:30:33",
+    example = "2022-08-24T10:30:33",
   )
-  @JsonFormat(pattern = "dd/MM/yyyy HH:mm:ss")
+  @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss")
```
andrewrlee, line 48: using standard ISO8601 dates

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/PublicLicenceController.kt
```diff
@@ -18,17 +18,17 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.resource.Tags
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.resource.publicApi.model.licence.Licence
 
 @RestController
-@RequestMapping("/public/licence", produces = [MediaType.APPLICATION_JSON_VALUE])
+@Tag(name = Tags.LICENCES)
+@PreAuthorize("hasAnyRole('VIEW_LICENCES')")
```
andrewrlee, line 22: Specify VIEW_LICENCES role.. we could extract out constants for this..

andrewrlee: protecting and tagging at controller level rather than each endpoint

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/AdditionalCondition.kt
```diff
@@ -0,0 +1,27 @@
...
+  @ArraySchema(
+    arraySchema = Schema(discriminatorProperty = "type"),
+    schema = Schema(
+      description = "Individual instances of this condition that appear on the licence. Most conditions can only appear on a licence once so would have a single instance but a subset can appear multiple times.",
+      implementation = AdditionalConditionInstance::class,
+    ),
+  )
+  val instances: List<AdditionalConditionInstance>,
+
+)
```
andrewrlee, line 27: two different types of instances using a discriminator so we can only add enumerate the list of electronic monitoring types to specific types of instances - the swagger ui does display this very well due to a very long running bug. 
It produces the correct schema and I've tweaked things so it looks like the standard condition instance type.

Not sure whether instances are making this unnecessary complex.. its slightly odd to the client that they can have a list of instances which will be either standard instances or electronic monitoring instances.

If we just had a flat list of (possibly duplicated) conditions we could have a union of types at the condition level- have a `electronic monitoring condition` type and `standard condition type`..
 
Would end up with some duplicated conditions for MEZ, etc.. though

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/AdditionalConditionInstance.kt
```diff
@@ -1,19 +1,83 @@
...
+) : AdditionalConditionInstance
+
+enum class ElectronicMonitoringType {
+  EXCLUSION_ZONE,
+  CURFEW,
+  LOCATION_MONITORING,
+  ATTENDANCE_AT_APPOINTMENTS,
+  ALCOHOL_MONITORING,
+  ALCOHOL_ABSTINENCE,
+}
```
andrewrlee, line 83: The types of EM will become part of the API contract which is a bit more robust than strings

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/LicenceSummary.kt
```diff
@@ -0,0 +1,68 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.resource.publicApi.model.licence
+
+import com.fasterxml.jackson.annotation.JsonFormat
```
andrewrlee, line 3: Adding summary endpoint which is basically: `Licence - Conditions`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/Licence.kt
```diff
@@ -44,28 +43,28 @@ data class Licence(
...
 
   @Schema(description = "The username of the person who last updated the licence", example = "X34433")
   val updatedByUsername: String,
 
-  @Schema(description = "The date and time that the licence was last updated", example = "24/08/2022 09:30:33")
-  @JsonFormat(pattern = "dd/MM/yyyy HH:mm:ss")
-  val updatedDate: LocalDateTime,
+  @Schema(description = "The date and time that the licence was last updated", example = "2022-08-24T12:30:33")
+  @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss")
+  val updatedDateTime: LocalDateTime,
```
justin-gardiner, line 63: this is nullable in the database and seems be null in prod for quite a few licences. I'm not sure how as it is set when you create the licence and I can't see where it gets nulled.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/LicenceSummary.kt
```diff
@@ -0,0 +1,68 @@
...
+  val bookingId: Long,
+
+  @Schema(description = "The case reference number (CRN) for the person on the licence", example = "X12444")
+  val crn: String,
+
+  @Schema(
+    description = "The username of the person who approved the licence on behalf of the prison governor",
+    example = "X33221",
+  )
+  val approvedByUsername: String,
```
justin-gardiner, line 41: can't remember if we were returning only active licences or not. If not then approvedBy fields could be null

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/Licence.kt
andrewrlee: it makes sense conceptually for it to be null. I'll update this

justin-gardiner (approved):
lgtm

PR #336, opened 2023-10-12
CVSL-1278 create endpoint to get licences based on prison number

gurps-singh:
This PR is to add a new endpoint to the CVL public API to allow a list of licence summary representations to be returned based on prison number.


justin-gardiner (approved):
lgtm

PR #341, opened 2023-10-20
CVSL-1425 Adding missing call to send reapproval email on licence edit

SebNegahbanMoJ:
By this time the code reaches the point of the added line, the only possibility is for the original licence to be APPROVED and the new edit to be IN_PROGRESS, therefore no additional checks are needed.


gurps-singh (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -593,6 +593,8 @@ class LicenceService(
     }
 
     val licenceCopy = copyLicenceAndConditions(licenceEntity, IN_PROGRESS)
+    notifyReApprovalNeeded(licenceEntity)
```
andrewrlee, line 596: This could do with a test?

gurps-singh (approved):

PR #343, opened 2023-10-20
CVSL-1162 - Updating curfew condition config

SebNegahbanMoJ:


gurps-singh:
Is there any tests around the policy - if we are adding to the test data json then is there somewhere we can check the structure of the json returned?

SebNegahbanMoJ:
I'm not sure if I'm misunderstanding your question, but what I think you're asking is whether there's a comparison between the Kotlin and Json representations of the policy docs?
If that's your question, then yes, there is! There's an integration test called `LicencePolicyIntegrationTest` which compares the kt file to the json, in other words testing the serlialisation of the kt to what will be returned via the licence policy endpoint.

In practice, it's meant to be that the JSON is the 'known correct' structure of the policy doc, and the integration test makes sure that the kt serialises correctly to form that json.

gurps-singh (approved):

justin-gardiner (approved):
lgtm

andrewrlee (approved):
LGTM

PR #344, opened 2023-10-23
Cvsl 1238 friday release change weekend bank hols alert

nishanthmahasamudram:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/ExclusionZoneIntegrationTest.kt
```diff
@@ -27,6 +30,7 @@ class ExclusionZoneIntegrationTest : IntegrationTestBase() {
   @BeforeEach
   fun setupClient() {
     webTestClient = webTestClient.mutate().responseTimeout(Duration.ofSeconds(60)).build()
+    govUkApiMockServer.stubGetBankHolidaysForEnglandAndWales()
```
andrewrlee, line 33: is this needed?

nishanthmahasamudram: Yes, we need it.
Because we are using ".uri("/licence/id/2")" this api which needs bankHoliday Service to return list of bank holidays'

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
```diff
@@ -187,4 +188,39 @@ data class Licence(
...
+  /** Friday is also considered as weekend */
+  private fun isBankHolidayOrWeekend(listOfBankHolidays: List<LocalDate>, releaseDate: LocalDate?): Boolean {
+    val dayOfWeek = releaseDate?.dayOfWeek
+    if (dayOfWeek == DayOfWeek.FRIDAY || dayOfWeek == DayOfWeek.SATURDAY || dayOfWeek == DayOfWeek.SUNDAY) {
+      return true
+    }
+    return listOfBankHolidays.contains(releaseDate)
+  }
+
+  private fun getEarliestReleaseDate(
```
andrewrlee, line 211: I think it'd make sense to put this on the BankHoliday service. 

in the licence service you could have something like:
```kt
val releaseDate = licence.actualReleaseDate ?: licence.conditionalReleaseDate
val earliestReleaseDate = bankHoliday.getEarliestReleaseDate(releaseDate) 
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
```diff
@@ -187,4 +188,39 @@ data class Licence(
...
+    possibleReleaseDates: MutableList<LocalDate>,
+  ): LocalDate {
+    val earlyPossibleDate = normalReleaseDate?.minusDays(1)
+    if (possibleReleaseDates.size < workingDays) {
+      if (earlyPossibleDate !== null && !isBankHolidayOrWeekend(listOfBankHolidays, earlyPossibleDate)) {
+        possibleReleaseDates.add(earlyPossibleDate)
+      }
+      getEarliestReleaseDate(listOfBankHolidays, workingDays, earlyPossibleDate, possibleReleaseDates)
+    }
+    return possibleReleaseDates[workingDays - 1]
```
andrewrlee, line 224: I'm not sure your type sig is correct - how can you return a non nullable date if your release date is nullable?

rather than using recursion, you could use [kotlin magic](https://kotlinlang.org/docs/sequences.html#from-a-function) to model it as a stream of dates..

something like:
```
val releaseDates = generateSequence(releaseDate) { it.minusDays(1) }
    .filterNot { isBankHoliday(it) }
    .take(workingDays)
    .reverse()
    .firstOrNull()
```

Not sure that logic is quite right... but hopefully close to what you want?

nishanthmahasamudram: If release date is null I am returning it immediately. If not only then we are calling getEarliestReleaseDate method.

src/main/resources/application.yml
```diff
@@ -122,6 +122,8 @@ notify:
 # Overridden in production
 internalEmailAddress: 'testemail@probation.gov.uk'
 
+workingDays: 3
```
andrewrlee, line 125: maybe a more descriptive name here? 
`maxNumberOfWorkingDaysAllowedForEarlyRelease`?

andrewrlee (approved):
LGTM, little comment but could be done in a follow up PR?

PR #350, opened 2023-10-27
CVSL-1437 Patching and adding missing transactional annotations

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -631,6 +634,7 @@ class LicenceService(
     licenceRepository.saveAndFlush(updatedLicenceEntity)
   }
 
+  @Transactional
```
andrewrlee, line 637: this was the only thing that actually does multiple updates and was called directly by a controller

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -958,16 +962,13 @@ class LicenceService(
...
     val licenceIds = licences.map { it.id }
     val licencesToDeactivate =
       licenceRepository.findAllByVersionOfIdInAndStatusCodeIn(licenceIds, listOf(IN_PROGRESS, SUBMITTED))
     if (licencesToDeactivate.isNotEmpty()) {
-      inactivateLicences(
-        licencesToDeactivate,
-        reason,
-        false,
-      )
+      inactivateLicences(licencesToDeactivate, reason, false)
```
andrewrlee, line 971: This was moaning that it was not going via the transactional proxy - it was only being called by transactional methods which I assumed would be fine but to remove doubt I made this method public/transactional

PR #351, opened 2023-10-27
CVSL-965: Add a partial index to licence table to prevent duplicate i…

justin-gardiner:
…n progress licences for a nomis id.


justin-gardiner:
********* NOT TO BE MERGED ************

This is just an example of how using a partiall index could be used to prevent more than one in progress licence being created for a nomis id.

PR #354, opened 2023-10-30
CVSL-1443 Reduce spam to govuk bankholiday endpoint

andrewrlee:
Recent changes mean that we are bypassing our cache when requesting bankholiday. This is currently happening twice per licence load.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/BankHolidayService.kt
```diff
@@ -1,34 +1,14 @@
...
-  @Value("\${maxNumberOfWorkingDaysAllowedForEarlyRelease}") private val maxNumberOfWorkingDaysAllowedForEarlyRelease: Int,
 ) {
 
   @Cacheable("bank-holidays")
   fun getBankHolidaysForEnglandAndWales() = govUkApiClient.getBankHolidaysForEnglandAndWales()
-
-  /** Friday is also considered as weekend */
-  fun isBankHolidayOrWeekend(releaseDate: LocalDate?): Boolean {
-    val listOfBankHolidays: List<LocalDate> = getBankHolidaysForEnglandAndWales()
-    val dayOfWeek = releaseDate?.dayOfWeek
```
justin-gardiner, line 22: so the issue was because it was calling a method in the same class?

justin-gardiner (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/BankHolidayService.kt
andrewrlee: yep that's right

PR #355, opened 2023-11-02
CVSL-1412 Add eligibility criteria for CVL to filter the search results

gurps-singh:
This PR introduces the eligibility service which will be used to check whether a person is eligible for the CVL service. This is introduced due to some support issues where users were searching and retrieving records using search which were ineligible for CVL. We now filter out the results from the prisoner search response where they are not eligible for CVL (i.e. released) based on the eligibility rules for using CVL from the front end.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -172,4 +183,77 @@ class ComService(
...
+      }
+    }
+
+    // Person has no CRD and a post recall release date - Recall case - not eligible for CVL
+    if (this.confirmedReleaseDate != null && this.conditionalReleaseDate != null && this.postRecallReleaseDate != null) {
+      return false
+    }
+
+    // Person is approved for HDC - not eligible for CVL
+    val hdcStatus = prisonApiClient.getHdcStatus(this.bookingId).block()
```
andrewrlee, line 252: mmh.. that's a hefty call to make for each licence.. good to have it at the end, maybe there's a bulk endpoint we can use instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -172,4 +183,77 @@ class ComService(
...
+
+    // Person's ARD or CRD is in the past - not eligible for CVL
+    if (this.confirmedReleaseDate != null) {
+      if (this.confirmedReleaseDate.isBefore(LocalDate.now()) || this.conditionalReleaseDate.isBefore(LocalDate.now())) {
+        return false
+      }
+    }
+
+    // Person has no CRD and a post recall release date - Recall case - not eligible for CVL
+    if (this.confirmedReleaseDate != null && this.conditionalReleaseDate != null && this.postRecallReleaseDate != null) {
```
andrewrlee, line 247: this doesn't match the description? should it be `== null`

gurps-singh: Changed

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -172,4 +183,77 @@ class ComService(
...
+    // Person has a status other than ACTIVE x or INACTIVE TRN - not eligible for CVL
+    if (this.status != null) {
+      if (!(this.status.startsWith("ACTIVE") || this.status == "INACTIVE TRN")) {
+        return false
+      }
+    }
+
+    // Person's ARD or CRD is in the past - not eligible for CVL
+    if (this.confirmedReleaseDate != null) {
+      if (this.confirmedReleaseDate.isBefore(LocalDate.now()) || this.conditionalReleaseDate.isBefore(LocalDate.now())) {
```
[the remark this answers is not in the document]
gurps-singh: Confirmed with Seb, if no confirmed release date and a CRD in the past - not eligible.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -0,0 +1,104 @@
...
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonerSearchPrisoner
+import java.time.Clock
+import java.time.LocalDate
+
+@Service
+class EligibilityService(
+  private val prisonApiClient: PrisonApiClient,
+  private val clock: Clock,
+) {
+
```
gurps-singh, line 14: Used && chaining logic due to the short circuit nature of the && call.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -0,0 +1,104 @@
...
+  private fun hasConditionalReleaseDate(conditionalReleaseDate: LocalDate?): Boolean {
+    return conditionalReleaseDate != null
+  }
+
+  private fun isOnEligibleExtendedDeterminateSentence(
+    paroleEligibilityDate: LocalDate?,
+    conditionalReleaseDate: LocalDate?,
+    actualReleaseDate: LocalDate?,
+    actualParoleDate: LocalDate?,
+  ): Boolean {
```
gurps-singh, line 58: PED null check left in after conversation with Seb.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+    return false
+  }
+
+  private fun hasEligibleReleaseDate(actualReleaseDate: LocalDate?, conditionalReleaseDate: LocalDate?): Boolean {
+    val releaseDate = actualReleaseDate ?: conditionalReleaseDate
+
+    return releaseDate!!.isEqual(LocalDate.now(clock)) || releaseDate.isAfter(LocalDate.now(clock))
+  }
+
```
gurps-singh, line 94: Due to chaining logic, we already know by the time we get here, a CRD exists. Omitted the use of the recall flag from the prisoner search due to inaccurate nature of it.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -0,0 +1,109 @@
...
+    if (postRecallReleaseDate == null) {
+      return false
+    }
+    // If the PRRD > CRD - it should be treated as a recall otherwise it is not treated as a recall
+    return postRecallReleaseDate.isAfter(conditionalReleaseDate!!)
+  }
+
+  private fun isHomeDetentionCurfewCase(bookingId: String, homeDetentionCurfewEligibilityDate: LocalDate?): Boolean {
+    val bookingIdRequest = bookingId.toLong()
+    val hdcStatus = prisonApiClient.getHdcStatus(bookingIdRequest).block()
```
andrewrlee, line 106: we need to limit the calls to prison api - if you consider that this could be called for every otherwise prisoner eligible.

I'd suggest moving this out of the eligibility service and using `getHdcStatuses` instead. 
Also it's only worth doing it for people who have a `homeDetentionCurfewEligibilityDate`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
andrewrlee: Not sure I like the idea of relying that this method was called after another method? would it be better to include it the clause?

Worth adding a comment about unreliability of recall flag to the source code?

These would probably be more nicely expressed as extension methods?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -0,0 +1,109 @@
...
+    if (actualParoleDate != null) {
+      return false
+    }
+
+    return true
+  }
+
+  private fun hasActivePrisonStatus(status: String?): Boolean {
+    if (status != null) {
+      if (status.startsWith("ACTIVE") || status == "INACTIVE TRN") {
```
andrewrlee, line 82: Maybe something like:
```kt
    if (status != null) {
      return status.startsWith("ACTIVE") || status == "INACTIVE TRN")
    }
   return false
```

or:
```kt
private fun hasActivePrisonStatus(status: String?) = status?.let { it.startsWith("ACTIVE") || it == "INACTIVE TRN" } ?: false
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
andrewrlee: would be tempted to use extension functions for these... would add a lot of uniformity.

The nice thing is you'd just have to remove the parameters from the functions and whack a `PrisonerSearchPrisoner.` to the method and it should all still compile

andrewrlee:
This is going to require some gnarly testing in preprod I think!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
gurps-singh: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityServiceTest.kt
```diff
@@ -0,0 +1,222 @@
...
+      aPrisonerSearchResult.copy(
+        paroleEligibilityDate = LocalDate.now(clock).minusDays(1),
+        actualParoleDate = LocalDate.now(clock).plusDays(1),
+      ),
+    )
+    assertThat(result).isFalse()
+  }
+
+  @Test
+  fun `Person is on ineligible EDS - has a APD with a PED in the past - not eligible for CVL `() {
```
SebNegahbanMoJ, line 107: I think this test may be a duplicate of the one above

SebNegahbanMoJ (dismissed):
Think there's a duplicate spec but non-blocking as it won't do any harm
Can be remove in a future PR if you'd prefer

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityServiceTest.kt
gurps-singh: Changed

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityServiceTest.kt
```diff
@@ -104,10 +104,21 @@ class EligibilityServiceTest {
...
-        paroleEligibilityDate = LocalDate.now(clock).minusDays(1),
+        paroleEligibilityDate = LocalDate.now(clock),
+        actualParoleDate = LocalDate.now(clock),
+      ),
+    )
+    assertThat(result).isFalse()
+  }
+
+  @Test
+  fun `Person is on ineligible EDS - has a APD with a PED in the future - not eligible for CVL `() {
```
SebNegahbanMoJ, line 118: No harm in having this, but I imagine it's actually return false on the future PED check earlier in the `isEligibleForCvl` check
Might be worth keeping regardless as it will protect against that PED check changing in future

SebNegahbanMoJ (approved):

PR #357, opened 2023-11-06
add /policy/latest endpoint

jaleen:
This PR adds a new endpoint /policy/latest endpoint This new endpoint returns most recent version of policy. At the moment its going to return version 2.1. This PR assumes that there is at least one policy present always.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/publicApi/PublicLicencePolicyService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/publicApi/PublicLicencePolicyService.kt
@@ -16,4 +16,9 @@ class PublicLicencePolicyService(
 
     return policy.transformToPublicLicencePolicy()
   }
+
+  fun getLatestLicencePolicy(): PublicLicencePolicy {
+    val transformToPublicLicencePolicy = licencePolicyService.currentPolicy()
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/publicApi/PublicLicencePolicyService.kt
```diff
@@ -16,4 +16,9 @@ class PublicLicencePolicyService(
 
     return policy.transformToPublicLicencePolicy()
   }
+
```
gurps-singh, line 19: Not a biggie but can be shortened to `return licencePolicyService.currentPolicy().transformToPublicLicencePolicy()`

gurps-singh (commented):
Need to add an integration test for the route - `src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/publicApi/PublicLicencePolicyServiceIntegrationTest.kt`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/publicApi/PublicLicencePolicyService.kt
jaleen: fixed

gurps-singh (approved):

PR #368, opened 2023-11-14
CVSL-1412 Exposing reasons for eligibility failure

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -5,87 +5,96 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.Pris
...
-        return true
+
+  private fun isPersonParoleEligible(): EligibilityCheck = early@{
+    if (it.paroleEligibilityDate != null) {
+      if (it.paroleEligibilityDate.isAfter(LocalDate.now(clock))) {
+        return@early true
       }
     }
-    return false
+    return@early false
```
andrewrlee, line 43: These labels are needed to return from a lambda expression - could tidy these up using `when`s

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -5,87 +5,94 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.Pris
...
-  private fun PrisonerSearchPrisoner.hasEligibleReleaseDate(): Boolean {
-    val releaseDate = this.confirmedReleaseDate ?: this.conditionalReleaseDate
+  private fun hasActivePrisonStatus(): EligibilityCheck = { prisoner ->
+    prisoner.status?.let {
+      it.startsWith("ACTIVE") || it == "INACTIVE TRN"
+    } ?: false
+  }
 
-    return releaseDate!!.isEqual(LocalDate.now(clock)) || releaseDate.isAfter(LocalDate.now(clock))
+  private fun hasReleaseDateInTheFuture(): EligibilityCheck = early@{
```
andrewrlee, line 79: I renamed this from `hasEligibleReleaseDate` to  `hasReleaseDateInTheFuture` and made it not fire if it was missing a release date - so that could be picked up by a different rule

andrewrlee: may be we need another rule for hasNoReleaseDate?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -5,87 +5,94 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.Pris
...
+    val releaseDate = it.confirmedReleaseDate ?: it.conditionalReleaseDate ?: return@early true
```
gurps-singh, line 80: Shouldn't this be if there is no confirmed release date and no conditional release date return false as then they do not have a release date in the future?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
gurps-singh: If they don't have a CRD, they would be ineligible and fail that earlier check and so be ineligible so is there a reason to check the other dates if they don't have the main CRD date?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityServiceTest.kt
```diff
@@ -213,7 +213,7 @@ class EligibilityServiceTest {
         conditionalReleaseDate = null,
       ),
     )
-    assertThat(result).containsExactly("has no conditional release date")
+    assertThat(result).containsExactly("has no conditional release date", "release date in past")
```
SebNegahbanMoJ, line 216: I think this is a touch misleading as in the case that they have no ARD or CRD, we'd still get the second reason of 'release date in past', but I suppose that's somewhat unavoidable if we're checking all potential exclusion reasons at once rather than having a 1-by-1 hierarchy

gurps-singh: I think this ties in with Andy's comment re having an additional rule checking for `noReleaseDate` which we can add later.

gurps-singh (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityServiceTest.kt
SebNegahbanMoJ: Ah, yep, missed that!

PR #369, opened 2023-11-15
CVSL-1412 add page size to licence caseload call

gurps-singh:
This PR is to add the page size to the initial licence caseload search call we make when using search. This is off the back of initial investigation around search results following the addition of the Eligibility Service. There are too few results being returned and this seems like the culprit.


andrewrlee (approved):
LGTM

PR #373, opened 2023-11-16
CVSL-1431 Adding ability to skip condition input

SebNegahbanMoJ:
Also moves validation checks for condition completeness to the API


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionWithConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.AdditionalCondition
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy.AdditionalConditions
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.policy.IAdditionalCondition
+
+data class AdditionalConditionWithConfig(
+  val additionalCondition: AdditionalCondition,
+  val config: IAdditionalCondition,
+)
+
+fun mapConditionsToConfig(licenceConditions: List<AdditionalCondition>, policy: AdditionalConditions): List<AdditionalConditionWithConfig> {
```
andrewrlee, line 12: if you make policy a `Set<IAdditionalCondition>`, then you can just pass in `policy.allAdditionalConditions()`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionWithConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+  val policyAdditionalConditions = policy.ap + policy.pss
+  return licenceConditions.map {
+    val policyCondition = policyAdditionalConditions.find { pc -> it.conditionCode == pc.code }!!
+    AdditionalConditionWithConfig(it, policyCondition)
+  }
+}
+
+fun checkConditionsReadyToSubmit(licenceConditions: List<AdditionalCondition>, policy: AdditionalConditions): Map<String, Boolean> {
+  if (licenceConditions.isEmpty()) { return emptyMap() }
```
andrewrlee, line 21: don't need this guard, an empty list is handled naturally anyway

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionPss.kt
```diff
@@ -10,4 +10,5 @@ data class AdditionalConditionPss(
   val pssDates: Boolean? = null,
   override val inputs: List<Input>? = null,
   override val type: String? = null,
+  override val skipable: Boolean? = null,
```
andrewrlee, line 13: is null a sensible default? making this mandatory might be better, you could default to false?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionAp.kt
```diff
@@ -10,4 +10,5 @@ data class AdditionalConditionAp(
   override val categoryShort: String? = null,
   val subtext: String? = null,
   override val type: String? = null,
+  override val skipable: Boolean? = null,
```
andrewrlee, line 13: `skippable`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceConditionService.kt
```diff
@@ -100,7 +101,12 @@ class LicenceConditionService(
 
     auditService.recordAuditEventAddAdditionalConditionOfSameType(licenceEntity, newCondition)
 
-    return transform(newCondition)
+    val readyToSubmit = checkConditionsReadyToSubmit(
+      listOf(newCondition),
+      licencePolicyService.policyByVersion(licenceEntity.version!!).additionalConditions,
+    )[newCondition.conditionCode]!!
```
andrewrlee, line 107: could just add a helper for dealing with a single condition in `checkConditionsReadyToSubmit `

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionWithConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+  val conditionsWithConfig = mapConditionsToConfig(licenceConditions, policy)
+  return conditionsWithConfig.associate {
+    val enteredFields = it.additionalCondition.additionalConditionData.map { data -> data.dataField }
+    val inputEntered =
+      if (!it.config.requiresInput) {
+        true
+      } else {
+        it.config.getConditionInputs()!!.map { input -> input.name }.any { name -> enteredFields.contains(name) }
```
andrewrlee, line 29: is this right? "any input has a data value"? shouldn't it be "all inputs have a corresponding data value " ?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionWithConfig.kt
```diff
@@ -0,0 +1,33 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
```
andrewrlee, line 1: this is domain logic so could probably sit in the services folder

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionWithConfig.kt
SebNegahbanMoJ: Yeah, that's right
Until we have `optional` flags on inputs in the condition config, we don't have a reliable way to determine what's optional and what's not in the API. It's all controlled by the frontend validators. 
Because of that, there are instances where a condition can have some of its inputs populated and not others, and that's valid and should be able to be submitted.
On the flip-side, for conditions that don't have optional inputs, the data is all-or-nothing, you can't submit condition data without submitting all of the required data (due to the frontend validators) meaning that the only time that a condition can have some of its inputs populated and not others are when the others are optional.

I thought adding those optional flags were somewhat out of scope for this ticket, so didn't add them now.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionAp.kt
SebNegahbanMoJ: I _knew_ it looked wrong 😅

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionPss.kt
SebNegahbanMoJ: Happy to default to false, I didn't want to bog-down the config with unnecessary declarations of `skippable = false` because that should be the assumption!
I _think_ there was a reason I went with `null` but I'll double check my own logic and can always refactor to account for default `false`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionWithConfig.kt
SebNegahbanMoJ: Ah, I assume the not-null check in `mapConditionsToConfig` would complain if the list was empty, as I thought `find` would return null on an empty list
`val policyCondition = policyAdditionalConditions.find { pc -> it.conditionCode == pc.code }!!`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AdditionalConditionWithConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+
+fun checkConditionsReadyToSubmit(licenceConditions: List<AdditionalCondition>, policy: Set<IAdditionalCondition>): Map<String, Boolean> {
+  val conditionsWithConfig = mapConditionsToConfig(licenceConditions, policy)
+  return conditionsWithConfig.associate {
+    val enteredFields = it.additionalCondition.additionalConditionData.map { data -> data.dataField }
+    val inputEntered =
+      if (!it.config.requiresInput) {
+        true
+      } else {
+        it.config.getConditionInputs()!!.map { input -> input.name }.any { name -> enteredFields.contains(name) }
```
andrewrlee, line 26: think maybe a a short comment saying that this relies on the assumption that any input provided means its ready to submit. Could also call `inputEntered`, `readyToSubmit` ? 

(that variable makes more sense for the situation when a condition doesn't require input too)

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionWithConfigTest.kt
```diff
@@ -0,0 +1,236 @@
...
+      assertThat(
+        checkConditionReadyToSubmit(
+          anAdditionalConditionEntity.copy(additionalConditionData = emptyList()),
+          aPolicy.allAdditionalConditions(),
+        ),
+      ).isFalse()
+    }
+
+    @Test
+    fun `returns true for a condition that doesn't need inputs`() {
```
andrewrlee, line 124: missing a test for some inputs but not all?

andrewrlee (dismissed):
LGTM

PR #374, opened 2023-11-17
A public end point to get licence by licence id

jaleen:
Added a public endpoint  public/licences/id/674 to retrieve the public licence.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
```diff
@@ -247,4 +254,36 @@ data class Licence(
 
   @Schema(description = "If ARD||CRD falls on Friday/Bank holiday/Weekend then it is eligible for early release)")
   val isEligibleForEarlyRelease: Boolean = false,
-)
+) {
+  fun transformToPublicLicence(): uk.gov.justice.digital.hmpps.createandvaryalicenceapi.resource.publicApi.model.licence.Licence {
```
gurps-singh, line 258: This needs to be in the correct file - all the public API transforms live here 
: `uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/publicApi/ToModelTransformers.kt`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/Licence.kt
```diff
@@ -53,7 +53,7 @@ data class Licence(
 
   @Schema(description = "The date and time that the licence was first created", example = "2022-08-24T11:30:33")
   @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss")
-  val createdDateTime: LocalDateTime,
+  val createdDateTime: LocalDateTime?,
```
gurps-singh, line 56: So this value is used in multiple places and should always be set, I believe. Making it nullable for some edge case?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToResourceTransformers.kt
```diff
@@ -0,0 +1,99 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
```
gurps-singh, line 1: Just slightly confused as to what this transform is doing in comparison to the transform that you currently have in the Licence. Should they be in one file?

run-local-with-dev-db.sh
```diff
@@ -0,0 +1,43 @@
+#
```
gurps-singh, line 1: Is this your personal script - don't think we need it committing?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/publicApi/PublicLicenceServiceTest.kt
```diff
@@ -589,5 +607,199 @@ class PublicLicenceServiceTest {
...
+      createdDateTime = LocalDateTime.of(2023, 10, 11, 11, 0, 0),
+      updatedByUsername = "TestUpdater",
+      updatedDateTime = LocalDateTime.of(2023, 10, 11, 11, 30, 0),
+      isInPssPeriod = false,
+      conditions = licenceConditions,
+
+    )
+  }
+
+  /****************/
```
gurps-singh, line 660: This needs to be above the companion object

run-local-with-dev-db.sh
jaleen: it felt useful :)

andrewrlee: I think this could be useful, it could possibly have been an extension of the other script. 

The comment doesn't match what it does though.. it would be good to fix that and we could maybe look at consolidating the two at a later date.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
andrewrlee: ideally we wouldn't need to mock the transformer functions, they're just pure functions.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToResourceTransformers.kt
```diff
@@ -0,0 +1,99 @@
...
+
+fun transformElectronicMonitoring(model: AdditionalCondition): ElectronicMonitoringAdditionalCondition {
+  return ElectronicMonitoringAdditionalCondition(
+    category = model.category.orEmpty(),
+    type = ConditionTypes.ELECTRONIC_MONITORING,
+    id = model.id ?: 0,
+    code = model.code.orEmpty(),
+    text = model.text.orEmpty(),
+    electronicMonitoringTypes = model.data.filter { data -> data.field == ELECTRONIC_MONITORING_TYPES }
+      .map { data -> ElectronicMonitoringType.find(data.value.orEmpty())!! },
```
andrewrlee, line 63: I think if we're `!!`ing them here, maybe it'd be better for `find` to return non-nullable and throw an exception if not present with some info about the thing it couldn't find?
in kotlin you can do:
```kt
someNullable ?: error("some message about the $thing")

File: src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToResourceTransformers.kt [hunk 7]
```diff
@@ -0,0 +1,99 @@
...
+// Transform a list of model additional conditions to resource additional conditions
+fun List<AdditionalCondition>.transformToResourceAdditional(): List<ModelAdditionalCondition> {
+  val electronicMonitoringConditions =
+    filter { condition -> condition.code == ELECTRONIC_TAG_COND_CODE }.map(::transformElectronicMonitoring)
+  val multipleExclusionZoneAdditionalConditions =
+    filter { condition -> condition.code == EXCLUSION_ZONE_COND_CODE }.map(::transformMultipleExclusionZonesCondition)
+  val stdConditions =
+    filter { condition -> condition.code != ELECTRONIC_TAG_COND_CODE && condition.code != EXCLUSION_ZONE_COND_CODE }.map(
+      ::standardAdditionalCondition,
+    )
```
andrewrlee, line 50: you could use a single `map` and a `when` here
```kt
when (condition.code) {
    ELECTRONIC_TAG_COND_CODE -> transformElectronicMonitoring(condition)
    EXCLUSION_ZONE_COND_CODE -> transformMultipleExclusionZonesCondition(condition)
    else -> standardAdditionalCondition(condition
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToResourceTransformers.kt
andrewrlee: think this should sit in `package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.publicApi`
Could maybe call it `LicenceDetailTransformer`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/Codes.kt
```diff
@@ -0,0 +1,4 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util
```
andrewrlee, line 1: could we move this to `uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies`
and update the usages inside that package to use these too?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
jaleen: yup, but these were not under test and were a dependency :)

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/publicApi/PublicLicenceServiceTest.kt
```diff
@@ -589,5 +608,197 @@ class PublicLicenceServiceTest {
...
+    )
+  }
+
+  @Nested
+  inner class `Get licence by id` {
+    @Test
+    fun `service returns a licence by id`() {
+      val licenceId: Long
+      licenceId = 12345
+      doReturn(pubLicence).whenever(modelLicenceMock).transformToPublicLicence()
```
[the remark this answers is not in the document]
jaleen: transformToPublicLicence is tested somewhere else. here testing the single line of code in PublicLicenceService which invokes transformToPublicLicence method on a dependency

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
andrewrlee: I'm not sure I agree that they are a dependency as such.. in kotlin they are resolved to static invocations. 
They're just pure functions with no side effects and there's not really any harm calling them directly in tests.
   
I think the reason why it mocking didn't work is that top level functions are static methods and need to mocked like [this](https://notwoods.github.io/mockk-guidebook/docs/mocking/extension/#:~:text=Kotlin%20lets%20you%20write%20functions,can%20be%20mocked%20using%20mockkStatic%20.)

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/publicApi/PublicLicenceServiceTest.kt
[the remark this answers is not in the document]
andrewrlee: by doing something like 
`assertThat(transformToPublicLicence (actualLicence)).isEqualTo(pubLicence)`

You'd be testing the behaviour of the service method without relying on the implementation of the transformer.
You're saying "this method returns a transformed object" without having to know any details about the transformation itself.. and you wouldn't need to worry about having to set up stubbing behaviour or creating an additional stubbed object

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
jaleen: fixed by using the actual licence object

andrewrlee: are we able to get rid of this now? can you just use the version in transformers?

andrewrlee (approved):
LGTM

PR #378, opened 2023-11-22
CVSL-1433 CVL Amend rules Nomis overrides take precedence

nishanthmahasamudram:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceFunctionsKtTest.kt
```diff
@@ -83,6 +89,78 @@ class LicenceFunctionsKtTest {
...
+        pedChanged = false,
+        npdChanged = false,
+        prrdChanged = false,
+        isMaterial = true,
+      ),
+    )
+  }
+
+  @Test
+  fun `Sentence Changes should return material change for SED when paroleEligibilityDate is updated`() {
```
SebNegahbanMoJ, line 101: I think there's a bit of a misnomer here, as the name of each of the new tests references `should return material change for SED`, but `SED` isn't changing in any of these tests

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceFunctionsKtTest.kt
```diff
@@ -83,6 +89,78 @@ class LicenceFunctionsKtTest {
...
+        isMaterial = true,
+      ),
+    )
+  }
+
+  @Test
+  fun `Sentence Changes should return material change for PRRD when postRecallReleaseDate is updated`() {
+    val licence = testLicence.copy(statusCode = LicenceStatus.APPROVED)
+
+    Assertions.assertThat(
```
andrewrlee, line 150: would be good to get into the habit of statically importing these

nishanthmahasamudram:
As per Andy's comments(Lets hold off setting the other dates in the database- as we would really need to do some kind of migration to back fill existing records which would be loads of work.) we are not adding PRRD, PED, or NPD to Licence table now and not reading them in both licence creation and data change events.

PR #387, opened 2023-12-06
CVSL-1505 add probation search call to api client

gurps-singh:
This PR is to add a `/search` call from the Probation Search API client as this is needed to move the licence creation to the backend.


louisenorris (approved):
LGTM

PR #394, opened 2023-12-11
Cvsl 1483 While performing case search, add checks to confirm if prisoner qualify to create a licence.

jaleen:
Check all prisoners from probation search results (irrespective of no licence / licence) for eligibility first. NOMIS status check and the CRD/ARD future check is done for everyone. For offenders with a licence,If in prison (so not on probation) and have a licence in draft state (so the statuses such as IN_PROGRESS), then return in the search results
If on probation and have a licence in an “on probation” state (so the statuses such as ACTIVE, VARIATION_IN_PROGRESS etc), then return in the search results.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceEventType.kt
```diff
@@ -15,6 +15,9 @@ enum class LicenceEventType {
   VARIATION_APPROVED,
   INACTIVE,
   RECALLED,
+  OOS_RECALL,
   VERSION_CREATED,
   NOT_STARTED,
+  OOS_BOTUS,
+  NOT_IN_PILOT,
```
andrewrlee, line 22: these aren't really types of event - they are technically statuses that are not supported
If you're not using them would suggest removing this

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -153,31 +158,57 @@ class ComService(
...
+    val prisonNumbers = record
+      .mapNotNull { (result, _) -> result.identifiers.noms }
+
+    val prisoners = this.prisonerSearchApiClient.searchPrisonersByNomisIds(prisonNumbers)
+      .filter { prisoner ->
+        eligibilityService.isEligibleForCvl(prisoner) ||
+          record.isLicenceInGivenStatusPresent(prisoner.prisonerNumber, listOf(LicenceStatus.ACTIVE))
+      }
+
+    if (prisoners.isEmpty()) {
```
andrewrlee, line 202: is this needed? if prisoners is empty you could just return it?

SebNegahbanMoJ:
If I'm understanding the ticket correctly, I think this refactor might be a bit heavy for the changes required.
I believe we could get away with only adding a check before [the call to createRecord](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/394/files#diff-823b9fc980e42d49d1c869342ec91cb90338b9a6c907b437c12c5c5bbda7d2c4L114) that does:

`if(!licence.isOnProbation(){ <check CRD is in the future> <check nomis record status is ACTIVE or INACTIVE_TRN> }`?

In other words, once we know they have an already-existing pre-ACTIVE licence, run the checks listed in the ticket.
While I agree conceptually with running all of the eligibility checks over cases with pre-ACTIVE licences, that's a wider discussion that we need to have with UCD, as we've already had reports of PPs getting confused why licences suddenly disappear from their caselists. If we add more reasons for these existing licences to disappear (such as HDC eligibility), I expect we'd get an uptick in support requests.
We can have this conversation before merging this PR if you'd like, but if we do want to run the eligibility checks over pre-ACTIVE licences too, I think that would be even simpler, as we could change the current eligibility check [here](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/394/files#diff-823b9fc980e42d49d1c869342ec91cb90338b9a6c907b437c12c5c5bbda7d2c4L158) to check cases that have: 
`{ licence == null || !licence.isOnProbation }`
This would subject the cases with licences in the create-side to the same checks as those without licences.

(nb - In my first code block, it may need to be `licence.isOnProbation == false` if there's a valid reason for that to ever be `null`, but I assume it's always defined.)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -153,31 +158,54 @@ class ComService(
...
-      .mapNotNull { (result, _) -> result.identifiers.noms }
-
-    val prisoners = this.prisonerSearchApiClient.searchPrisonersByNomisIds(prisonNumbers)
+  private fun findPrisonersWithoutDraftLicences(
+    eligiblePrisonersList: List<PrisonerSearchPrisoner>,
+    record: List<Pair<CaseloadResult, Licence?>>,
+  ): Map<String, PrisonerSearchPrisoner> {
+    val eligiblePrisoners =
+      eligiblePrisonersList.filter { prisoner ->
+        !record.isLicenceInGivenStatusPresent(prisoner.prisonerNumber, LicenceStatus.DRAFT_LICENCES) ||
```
gurps-singh, line 167: Aren't these checks the inverse of each other?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -153,31 +158,54 @@ class ComService(
...
+      .filter { prisoner ->
+        eligibilityService.isEligibleForCvl(prisoner) ||
+          record.isLicenceInGivenStatusPresent(prisoner.prisonerNumber, listOf(LicenceStatus.ACTIVE))
+      }
+
+    return prisoners
+  }
+
+  private fun filterPrisonersWithNonDraftLicences(record: List<Pair<CaseloadResult, Licence?>>): List<Pair<CaseloadResult, Licence?>> =
+    record.filter { (_, licence) -> licence == null || licence.statusCode in LicenceStatus.DRAFT_LICENCES || licence.statusCode in LicenceStatus.ON_PROBATION_STATUSES.toList() }
```
gurps-singh, line 206: This check is duplicated as we are already checking "DRAFT_LICENCES" and "ON_PROBATION_STATUSES" previously?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -103,14 +103,19 @@ class ComService(
       body.getSortBy(),
     )
 
-    val resultsWithLicences: List<Pair<CaseloadResult, Licence?>> =
+    val prisonersPlusLicencesList: List<Pair<CaseloadResult, Licence?>> =
```
andrewrlee, line 106: CaseLoadResults are the result of a search against probation, we often call these offenders rather than prisoners (not everyone in probation is in prison) so this is slightly confusing name.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceStatus.kt
```diff
@@ -13,25 +13,22 @@ enum class LicenceStatus {
   VARIATION_REJECTED,
   VARIATION_APPROVED,
   NOT_STARTED,
+  OOS_RECALL,
+  OOS_BOTUS,
+  NOT_IN_PILOT,
```
andrewrlee, line 18: we can remove these - there not actually set anywhere.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -153,31 +158,54 @@ class ComService(
...
+    }
+
+  private fun findPrisonersEligibleToCreateLicence(record: List<Pair<CaseloadResult, Licence?>>): List<PrisonerSearchPrisoner> {
+    val prisonNumbers = record
+      .mapNotNull { (result, _) -> result.identifiers.noms }
+
+    val prisoners = this.prisonerSearchApiClient.searchPrisonersByNomisIds(prisonNumbers)
+      .filter { prisoner ->
+        eligibilityService.isEligibleForCvl(prisoner) ||
+          record.isLicenceInGivenStatusPresent(prisoner.prisonerNumber, listOf(LicenceStatus.ACTIVE))
```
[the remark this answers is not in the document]
andrewrlee: is eligible for cvl should only be applied for people who don't already have a licence
The new rules are only for people who haven't got a licence

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
```diff
@@ -103,14 +103,19 @@ class ComService(
...
       teamCaseloadResult.map { it to getLicence(it) }
 
-    val resultsWithUnstartedLicences = findPrisonersForUnstartedRecords(resultsWithLicences)
+    val eligiblePrisonersList = findPrisonersEligibleToCreateLicence(prisonersPlusLicencesList)
 
-    val searchResults = resultsWithLicences.mapNotNull { (result, licence) ->
+    val eligiblePrisonersWithoutDraftLicences =
+      findPrisonersWithoutDraftLicences(eligiblePrisonersList, prisonersPlusLicencesList)
+
+    val eligiblePrisonersForLicenceCreation = filterPrisonersWithNonDraftLicences(prisonersPlusLicencesList)
```
andrewrlee, line 114: So currently: 

1.) retrieve cases for probation and join with associated licences
2.)  findPrisonersEligibleToCreateLicence(1) 
3.)  findPrisonersWithoutDraftLicences(2, 1)
4.)  filterPrisonersWithNonDraftLicences(1)
5.)  4.map(   notStartedRecord(3) VS started())  // depending on whether licence is present

Maybe some thing like this would be better?

1.) search for cases for probation for this user/team 
2.) search for all prison records for 1.)
3.) join 1.) with any associated licences and prison records
4.) 3.filter(   isEligibleNotStartedRecord() VS isEligibleStartedRecord())  // depending on whether licence is present
5.) 4.map(   createNotStartedRecord() VS createStartedRecord()) // depending on whether licence is present

Now you always need to find the associated prisoner search details, doing that upfront might be simpler before filtering.

This would allow all the filtering logic to be in one place for each type of licence, and so a bit easier to read?

The only criteria/logic this ticket should be changing is by implementing something that encapsulates the concept of `isEligibleStartedRecord()`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ComService.kt
jaleen: yup, second condition is covered implicitly by first condition and probabation cases would be covered. so fixed it by removing second condition.

PR #396, opened 2023-12-13
CVSL-1504 extending com table to hold prison staff

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/CommunityOffenderManager.kt
```diff
@@ -1,33 +1,62 @@
...
 @Entity
-@Table(name = "community_offender_manager")
-data class CommunityOffenderManager(
-  @Id
-  @GeneratedValue(strategy = GenerationType.IDENTITY)
-  @NotNull
-  val id: Long = -1,
-
-  @Column(unique = true)
+@DiscriminatorValue(value = "COMMUNITY_OFFENDER_MANAGER")
```
andrewrlee, line 9: We now have a staff table with two subclasses and a discriminator

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/CrdLicence.kt
```diff
@@ -250,7 +250,7 @@ class CrdLicence(
 
   override fun deactivate() = copy(statusCode = LicenceStatus.INACTIVE)
 
-  override fun submit(submittedBy: CommunityOffenderManager) = copy(
+  fun submit(submittedBy: CommunityOffenderManager) = copy(
```
andrewrlee, line 253: have different submit methods depending on the type of licence

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/HardStopLicence.kt
```diff
@@ -0,0 +1,482 @@
...
+import jakarta.persistence.ManyToOne
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceKind
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceType
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+@Entity
+@DiscriminatorValue(value = "HARDSTOP")
+class HardStopLicence(
```
andrewrlee, line 16: This isn't actually being used anywhere yet, but has all the required fields.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/CommunityOffenderManagerRepository.kt
```diff
@@ -3,9 +3,10 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.repository
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.CommunityOffenderManager
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.Staff
 
 @Repository
-interface CommunityOffenderManagerRepository : JpaRepository<CommunityOffenderManager, Long> {
+interface CommunityOffenderManagerRepository : JpaRepository<Staff, Long> {
```
andrewrlee, line 9: I'll rename this in a follow up - it touches quite a few files

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -391,7 +392,12 @@ class LicenceService(
     val submitter = communityOffenderManagerRepository.findByUsernameIgnoreCase(username)
       ?: throw ValidationException("Staff with username $username not found")
 
-    val updatedLicence = licenceEntity.submit(submitter)
+    val updatedLicence = when (licenceEntity) {
+      is CrdLicence -> licenceEntity.submit(submitter)
+      is VariationLicence -> licenceEntity.submit(submitter)
+      is HardStopLicence -> TODO("Submitting hard stop licences not supported yet")
```
andrewrlee, line 398: should have some validation here - ensure correct type of user per licence and load up the correct DB record to associate them with

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceEventType.kt
```diff
@@ -7,6 +7,8 @@ enum class LicenceEventType {
   APPROVED,
   ACTIVATED,
   SUPERSEDED,
+  HARDSTOP_CREATED,
+  HARDSTOP_SUBMITTED,
```
andrewrlee, line 11: created two distinct events for these

src/main/resources/application.yml
```diff
@@ -58,7 +58,7 @@ spring:
       validationTimeout: 500
 
   flyway:
-    locations: classpath:/migration/common
+    locations: classpath:/migration/common,/migration/{vendor}
```
andrewrlee, line 61: make postgres migrations accessible when running against postgres

src/main/resources/migration/common/V10__create_com_details_table.sql
```diff
@@ -1,8 +1,8 @@
 CREATE TABLE community_offender_manager
 (
-    id SERIAL NOT NULL PRIMARY KEY,
-    staff_identifier INTEGER NOT NULL UNIQUE,
-    username VARCHAR(40) NOT NULL UNIQUE,
+    id SERIAL NOT NULL constraint community_offender_manager_key PRIMARY KEY,
```
andrewrlee, line 3: this is modifying an existing migration which is usually a big no no - the problem is that we can't modify the unique staff identifier index in a migration as there is no name to refer to in integration tests.

H2 in memory db generates a random name for each constraint if not explicitly named,  whereas postgres defaults to a name with a known format.

This change is explicitly naming the constraint to be the same as postgres has defaulted to in production - so is effectively a no-op that only enables us to write migrations that work in integration tests.

Usually flyway validates all migration scripts contents against what has been previously run to try and detect this kind of nefariousness before running any new migrations. 

To get round this I've added a prevalidate script that tweaks the checksum in the DB to the value for the newly modified script. As mentioned above - this is safe as we know that specifiying the name of the index to be what postgres has called it anyway has no effect.

src/main/resources/migration/postgresql/V33__delius-staff-id-unique-index.sql
```diff
@@ -0,0 +1,2 @@
+
+CREATE UNIQUE INDEX staff_staff_identifier_key ON staff(staff_identifier) WHERE KIND = 'COMMUNITY_OFFENDER_MANAGER';
```
andrewrlee, line 2: reinstating the unique index but only for `community_offender_manager`s

src/main/resources/migration/common/V32__support_storing_case_admins.sql
```diff
@@ -0,0 +1,25 @@
+ALTER TABLE community_offender_manager ADD COLUMN kind varchar(30);
+UPDATE community_offender_manager SET kind = 'COMMUNITY_OFFENDER_MANAGER';
+ALTER TABLE community_offender_manager ALTER COLUMN kind SET NOT NULL;
+
+ALTER TABLE community_offender_manager ALTER COLUMN staff_identifier DROP NOT NULL;
+ALTER TABLE community_offender_manager DROP CONSTRAINT community_offender_manager_staff_identifier_key;
+
+ALTER TABLE community_offender_manager RENAME TO staff;
```
andrewrlee, line 8: rename table to something generic

src/main/resources/migration/common/V32__support_storing_case_admins.sql
```diff
@@ -0,0 +1,25 @@
...
+
+ALTER TABLE licence ADD COLUMN submitted_by_ca_id INTEGER NULL REFERENCES staff(id);
+CREATE INDEX idx_licence_submitted_by_ca_id ON licence(submitted_by_ca_id);
+
+ALTER TABLE licence ADD COLUMN review_date TIMESTAMP WITH TIME ZONE;
+
+ALTER TABLE licence ADD COLUMN substitute_of_id INTEGER NULL REFERENCES licence(id);
+CREATE INDEX idx_licence_substitute_of_id ON licence(substitute_of_id);
+
+CREATE INDEX idx_licence_responsible_com_id ON licence(responsible_com_id);
```
andrewrlee, line 23: adding missing indexes for FKeys

src/main/resources/migration/postgresql/beforeValidate.sql
```diff
@@ -0,0 +1,26 @@
...
+language plpgsql
+$$
+BEGIN
+    IF EXISTS (select 1 from pg_class where relname='flyway_schema_history') THEN
+        UPDATE flyway_schema_history
+        SET checksum = -1711289223
+        WHERE (version, checksum) = ('10', -2012135289);
+    END IF;
+END;
+$$
```
andrewrlee, line 26: This has to be wrapped in a transaction and a test as we often tear down the DB and run all migrations when running locally. On an empty schema this runs before the flyway tables are created

src/main/resources/migration/postgresql/beforeValidate.sql
```diff
@@ -0,0 +1,26 @@
...
+This has no practical effect in production but enables us to write migrations that can be run in dev to modify the constraints by name.
+*/
+DO
+language plpgsql
+$$
+BEGIN
+    IF EXISTS (select 1 from pg_class where relname='flyway_schema_history') THEN
+        UPDATE flyway_schema_history
+        SET checksum = -1711289223
+        WHERE (version, checksum) = ('10', -2012135289);
```
andrewrlee, line 23: this rewrites the first migrations checksum so we can modify the contents to specify constraint names, as described above

src/main/resources/migration/common/V10__create_com_details_table.sql
gurps-singh: I saw modifying a migration and grimaced but use case is fair having explored alternatives. Would have been nice if it allowed you to add a constraint such as this as a separate migration such as:

`ALTER TABLE community_offender_manager ADD CONSTRAINT constraint community_offender_manager_key UNIQUE (id);`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/CommunityOffenderManager.kt
```diff
@@ -1,33 +1,62 @@
...
+) : Creator, Staff(
+  id = id,
+  kind = StaffKind.COMMUNITY_OFFENDER_MANAGER,
+  username = username,
+  email = email,
+  firstName = firstName,
+  lastName = lastName,
+  lastUpdatedTimestamp = lastUpdatedTimestamp,
+) {
+  fun copy(
```
jaleen, line 27: not sure if we want to consider the Lombok's Builder with copy constructors and use of With might reduce our boilerplate code

src/main/resources/migration/common/V32__support_storing_case_admins.sql
```diff
@@ -0,0 +1,25 @@
+ALTER TABLE community_offender_manager ADD COLUMN kind varchar(30);
+UPDATE community_offender_manager SET kind = 'COMMUNITY_OFFENDER_MANAGER';
+ALTER TABLE community_offender_manager ALTER COLUMN kind SET NOT NULL;
+
+ALTER TABLE community_offender_manager ALTER COLUMN staff_identifier DROP NOT NULL;
```
gurps-singh, line 5: Is this removing the staff_identifiers present in the table?

src/main/resources/migration/common/V10__create_com_details_table.sql
andrewrlee: you can.. but only if you know the name of the constraint which you don't in this case!  🤣

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/CommunityOffenderManager.kt
andrewrlee: There is a lot of boliler plate when not using data classes!

[lombok/kotlin](https://kotlinlang.org/docs/lombok.html) support is a bit experimental at the moment

I think we should probably embrace mutability a bit in the JPA code

src/main/resources/migration/common/V32__support_storing_case_admins.sql
andrewrlee: no - it drops the not null constraint
It definitely sounds like it could delete data so I double checked!

src/main/resources/migration/common/V32__support_storing_case_admins.sql
andrewrlee: yes, I think we need to do this out of hours

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceEventType.kt
jaleen, line 11: I am thinking aloud, if we need new types or existing types can suffice. In my head HardStop is still a licence and statuses are similar, its just being created/triggered by a different place. not sure. I don't see any real issues with two new status either though.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceKind.kt
```diff
@@ -19,4 +19,10 @@ enum class LicenceKind(
     LicenceEventType.VARIATION_CREATED,
     LicenceEventType.VARIATION_SUBMITTED,
   ),
+
+  HARDSTOP(
+    IN_PROGRESS,
+    LicenceEventType.HARDSTOP_CREATED,
+    LicenceEventType.HARDSTOP_SUBMITTED,
+  ),
```
jaleen, line 27: Its not necessarily a code change request just I am only thinking aloud. my feeling is that draft licence workflow/state might be generic, like create, amend, submit, approve, then separately it might be status of active, inactive irrespective of what type of licence it is. I appreciate some types of draft licence might have their unique states and rules as well.

jaleen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceKind.kt
andrewrlee: Yes, think the workflow states are pretty generic for licences, but think its worth being able to measure or track events specific to the type of licence for auditing purposes - tells a more compelling story when looking across licences for a given prisoner"
```
LICENCE_CREATED
LICENCE_TIMEDOUT
HARDSTOP_CREATED
HARDSTOP_SUBMITTED
LICENCE_ACTIVATED
VARIATION_CREATION
VARIATION_ACTIVATED
...
```

PR #399, opened 2023-12-14
CVSL-1505 add parameters needed to create licences in the backend

gurps-singh:
This latest PR introduces some extra variables we will need to expose from responses to various external API's in order to create the licence in the backend.


andrewrlee (approved):
LGTM

PR #401, opened 2023-12-15
CVSL-1514 Creating expire licences job in the API

SebNegahbanMoJ:
API work to move the expire-licences job from the frontend to the backend
Has the added benefit of fixing the current timeouts!


gurps-singh (commented):
I can see with the other jobs, there are accompanying unit tests - is this functionality covered in the integration tests instead? If not, this is missing the unit tests? See `uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/LicenceActivationServiceTest.kt`

gurps-singh (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
```diff
@@ -72,6 +72,16 @@ interface LicenceRepository : JpaRepository<Licence, Long>, JpaSpecificationExec
   """,
   )
   fun getAllVariedLicencesInPSSPeriod(): List<Licence>
+
+  @Query(
+    """
+    SELECT l
+        FROM Licence l
+        WHERE l.statusCode != uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.INACTIVE
```
andrewrlee, line 80: this looks like the wrong status? grabbing all inactive licences to inactivate?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/LicenceExpiryIntegrationTest.kt
```diff
@@ -0,0 +1,71 @@
...
+      }
+      .contains(
+        Tuple.tuple(2L, LicenceStatus.INACTIVE),
+        Tuple.tuple(5L, LicenceStatus.INACTIVE),
+        Tuple.tuple(6L, LicenceStatus.INACTIVE),
+      )
+
+    val remainingLicences = webTestClient.post()
+      .uri("/licence/match")
+      .bodyValue(MatchLicencesRequest(status = LicenceStatus.IN_FLIGHT_LICENCES))
```
andrewrlee, line 53: This doesn't look right

SebNegahbanMoJ: May I ask why? The purpose of this set of assertions is to make sure that the licences we'd expect to be unaffected are still in the statuses they were in originally. The easiest way to do so was to fetch all of the in-flight licences after the method-in-question had been run

PR #403, opened 2023-12-18
CVSL-1541 Changing ready-to-submit check to use condition policy version rather than licence policy version

SebNegahbanMoJ:
Fixes a bug that causes 500 errors for people going through the policy upgrade journey.


andrewrlee (dismissed):
LGTM

jaleen (approved):

andrewrlee (approved):
LGTM

PR #410, opened 2023-12-20
CVSL-1556 Adding conditional inputs to the check for ready-for-submission status for a licence

SebNegahbanMoJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AdditionalConditionWithConfig.kt
```diff
@@ -28,7 +28,8 @@ fun isLicenceReadyToSubmit(licenceConditions: List<AdditionalCondition>, policy:
       if (!it.config.requiresInput) {
         true
       } else {
-        it.config.getConditionInputs()!!.map { input -> input.name }.any { name -> enteredFields.contains(name) }
+        val policyInputs = it.config.getConditionInputs()!!.map { input -> input.getAllFieldNames() }.flatten()
```
andrewrlee, line 31: flatMap rather than flat and map

andrewrlee (approved):
LGTM

PR #417, opened 2024-01-03
Alee/cvsl 1505 fix issue with pss

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -47,6 +47,7 @@ class LicenceCreationService(
 
   companion object {
     private val log = LoggerFactory.getLogger(LicenceCreationService::class.java)
+    var FRONTEND_PAYLOAD_TAKES_PRIORITY = true
```
andrewrlee, line 50: bit of a short lived dirty hack until we know the new mechanism works. It allows us to easily test the new logic

SebNegahbanMoJ (approved):

PR #418, opened 2024-01-04
CVSL-1505 creating responsible COM in DB if they have not previously …

andrewrlee:
…logged in

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/User.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/User.kt
@@ -1,5 +1,14 @@
 package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation
 
 data class User(
+  val email: String?,
+  val username: String?,
+  val staffIdentifier: Long?,
+  val staff: StaffHuman?,
   val teams: List<Detail>,
+)
+
+data class StaffHuman(
+  val forenames: String?,
+  val surname: String?,
 )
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/User.kt
```diff
@@ -1,5 +1,14 @@
...
 
 data class User(
+  val email: String?,
+  val username: String?,
+  val staffIdentifier: Long?,
+  val staff: StaffHuman?,
   val teams: List<Detail>,
 )
+
+data class StaffHuman(
```
jaleen, line 11: how about StaffHuman->StaffName?

gurps-singh: Think this is what it is called in the CommunityApi?

gurps-singh (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/User.kt
andrewrlee: yes, the names are a bit rubbish in CommunityApi, also `User` refers to `Staff` which is really confusing as there is another entity called `User`!  Thought it better to match what the swagger exposes for consistency. 

Hopefully the probation integration team will replace this with a bespoke CVL API at some point.

jaleen (approved):

PR #419, opened 2024-01-04
CVSL-1504 Adding auditing to updating initial appointment information

SebNegahbanMoJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
```diff
@@ -269,14 +275,29 @@ class AuditService(
...
     val authUsername = SecurityContextHolder.getContext().authentication.name
-    val currentUser = staffRepository.findByUsernameIgnoreCase(authUsername)
+    val currentUser = if (userDetails != null) {
+      mapOf(
+        "username" to authUsername,
+        "firstName" to userDetails["forename"],
+        "lastName" to userDetails["surname"],
+      )
+    } else {
+      staffRepository.findByUsernameIgnoreCase(authUsername)?.let {
```
andrewrlee, line 288: this will only return COMs.. wondering if we need to pass the user details any more now we know the user will always exist in the DB.. I guess we didn't at the start of this ticket but do now?

SebNegahbanMoJ: The `else` branch of this should never be hit for initial appointment updates, as the `userDetails` should always be populated by having the `notBlank` constraint on the relevant requests.
The `else` branch exists for all other types of audits, as `createAuditEvent` is widely used. When we get the DB set up to house both prison and probation staff, we can likely revert much of this to just check our own DB rather than passing details through

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
```diff
@@ -271,6 +277,7 @@ class AuditService(
     changes: Map<String, Any>,
   ): AuditEvent {
     val authUsername = SecurityContextHolder.getContext().authentication.name
+    println(authUsername)
```
jaleen, line 280: do we want to print the username in logs?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
```diff
@@ -283,4 +290,24 @@ class AuditService(
       changes = changes,
     )
   }
+
+  private fun createUserAuditEvent(
```
jaleen, line 294: LGTM, I also like using Envers or @EnableJpaAuditing with @EntityListeners, JPA auditing have some limitations though

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
```diff
@@ -283,4 +290,24 @@ class AuditService(
...
+    licence: Licence,
+    summary: String,
+    changes: Map<String, Any>,
+  ): AuditEvent {
+    val authUsername = SecurityContextHolder.getContext().authentication.name
```
jaleen, line 299: not sure if we can use @LastModifiedBy annotation here by any chance?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
andrewrlee: removed this

jaleen (approved):

PR #420, opened 2024-01-05
CVSL-1515 add sqs library

gurps-singh:
This PR is to add the sqs library in preparation of adding logic to publish events from CVL.


louisenorris (approved):
LGTM

PR #428, opened 2024-01-12
CVSL-1515 access secret from cloud platform

gurps-singh:
This PR is to access the secret surfaced from Cloud Platform and to update our config with `serviceAccountName` to set the service account in CVL.


andrewrlee (approved):
LGTM

PR #429, opened 2024-01-15
Cvsl 1568 document api client

jaleen:


jaleen:
CVSL-1568 resolve conflicts

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/document/DocumentApiClient.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.core.io.ByteArrayResource
+import org.springframework.http.MediaType
+import org.springframework.http.client.MultipartBodyBuilder
+import org.springframework.stereotype.Service
+import org.springframework.web.reactive.function.client.WebClient
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.document.Document
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.document.DocumentMetaData
```
andrewrlee, line 12: the model package is for representations that we expose, so these two can be moved to the is package. 
They don't _need_ swagger docs.. but no harm adding them

PR #431, opened 2024-01-15
CVSL-1516 add localstack

gurps-singh:
This PR is to add localstack to CVL which will enable us to create mock versions of AWS services such as queues and notifications within our integration tests and locally, making tests queues, a lot easier.


louisenorris (approved):
LGTM

PR #433, opened 2024-01-16
CVSL-1516 add env var to values.yaml

gurps-singh:
This PR is to introduce the HMPPS_SQS_USE_WEB_TOKEN to our helm deploy as we are now using IRSA.


louisenorris (approved):
LGTM

PR #436, opened 2024-01-17
CVSL-1516 add extra config to change port numbers for localstack locally

gurps-singh:
This PR adds some changes to the `docker-compose.yml` file to allow two Localstack instances to run at the same time. Once we move events to the backend and remove Localstack from the front end, we can revert these values back to the default values.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,10 +24,16 @@ services:
       - hmpps
     container_name: localstack-api
     ports:
-      - "4566:4566"
+      - "4666:4666"
+      - "4610-4659:4610-4659"
     environment:
       - DEBUG=${DEBUG- }
       - DOCKER_HOST=unix:///var/run/docker.sock
+      - GATEWAY_LISTEN=0.0.0.0:4666
+      - EXTERNAL_SERVICE_PORTS_START=4610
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -24,10 +24,16 @@ services:
       - hmpps
     container_name: localstack-api
     ports:
-      - "4566:4566"
+      - "4666:4666"
```
andrewrlee, line 27: not sure this is correct? usually your mapping from an internal port to an external port?

gurps-singh: Followed the localstack docs for this?

https://docs.localstack.cloud/references/external-ports/#:~:text=If%20you%20wish%20to%20run,mapped%20to%20non%2Doverlapping%20ranges.

andrewrlee: could you just do something like 

> 8080:80 | Map port 8080 on the Docker host to TCP port 80 in the container.

So you could do `4666:4566` and that would make this containers internal 4556 port available as 4666?

andrewrlee (approved):
LGTM

docker-compose.yml
andrewrlee: ok, guess localstack is a bit more complicated

PR #440, opened 2024-01-22
CVSL-1587 Adding isInHardStopPeriod flag

SebNegahbanMoJ:
Defaulting to false until further development


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/Licence.kt
```diff
@@ -77,4 +77,7 @@ data class Licence(
 
   @Schema(description = "The AP and PSS conditions that form the licence")
   val conditions: Conditions,
+
+  @Schema(description = "Is the licence in the hard stop period? (Within two working days of release)")
+  val isInHardStopPeriod: Boolean,
```
andrewrlee, line 82: probably don't want it on the public api? bit of an internal impl detail

SebNegahbanMoJ: That's fair, I thought it may be helpful for external users to know when a CA is able to create the licence, but I suppose the main thing would be if the CA actually does, which we can expose in other ways

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/LicenceSummary.kt
```diff
@@ -76,4 +76,6 @@ data class LicenceSummary(
   @Schema(description = "Whether the licence in PSS period? This is when Licence End Date < TODAY <= TUSED (Top Up Supervision End Date)")
   val isInPssPeriod: Boolean,
 
+  @Schema(description = "Is the licence in the hard stop period? (Within two working days of release)")
+  val isInHardStopPeriod: Boolean,
```
andrewrlee, line 80: Not sure we need it here either

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -236,6 +237,7 @@ fun toVariation(
   licenceVersion = licence.licenceVersion,
   earliestReleaseDate = earliestReleaseDate,
   isEligibleForEarlyRelease = isEligibleForEarlyRelease,
+  isInHardStopPeriod = false,
```
andrewrlee, line 240: if it doesn't make sense to exist on a variation, don't need to add it to the subclass..

SebNegahbanMoJ: Hmm, I thought I tried to just excluded and it complained at me about it being required.

I must be mistaken, I'll remove it now

andrewrlee (approved):
LGTM

PR #442, opened 2024-01-23
CVSL-1517 add domain events service

louisenorris:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceOverrideService.kt
```diff
@@ -89,6 +95,46 @@ class LicenceOverrideService(
...
+        licence.id.toString(),
+        licence.crn,
+        licence.nomsId,
+      )
+    }
+
+    // if the current licence is a variation and the request status is inactivate - send domain event
+    if (licence.kind == LicenceKind.VARIATION && newStatus == LicenceStatus.INACTIVE) {
+      domainEventsService.recordDomainEvent(
+        LicenceDomainEventType.LICENCE_VARIATION_INACTIVATED,
```
andrewrlee, line 132: might be better to extract out a `getDomainEventType(licence.kind, newStatus)`?

good opportunity to use `when` and independently testable

andrewrlee: if there is no event type (`return null`) then you wouldn't need to send an event.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -201,6 +206,46 @@ class LicenceService(
...
+      domainEventsService.recordDomainEvent(
+        LicenceDomainEventType.LICENCE_ACTIVATED,
+        updatedLicence.id.toString(),
+        updatedLicence.crn,
+        updatedLicence.nomsId,
+      )
+    }
+
+    // if the current licence is a variation approved and the request status is active - send domain event
+    if (isCurrentLicenceVariationApproved && request.status == ACTIVE) {
```
andrewrlee, line 221: if you extracted out the helper method as described above you could use it here too?

(this is making the existing method very large in a very large class and there's a lot of boiler plate/duplication - in situations like this it's worth thinking of refactoring techniques that can help address this)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -468,6 +520,24 @@ class LicenceService(
             eventDescription = "${reason ?: "Licence automatically inactivated"} for ${licence.forename} ${licence.surname}",
           ),
         )
+
+        if (licence.kind == CRD) {
```
andrewrlee, line 524: could remove the conditionals/duplication here by adding  `inactivedDomainEventType` and `activatedDomainEventType` to LicenceKind. 

Need to support `HardStop` Licence types too which would be `LICENCE_ACTIVATED` / `LICENCE_INACTIVATED`

By adopting that pattern it enforces handling every licence type

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/DomainEventsService.kt
```diff
@@ -0,0 +1,73 @@
...
+    val personReferenceIdentifiers = PersonReference(
+      listOf(Identifiers("CRN", crn), Identifiers("NOMS", nomsNumber)),
+    )
+    val eventType = event.value
+    val occurredAt = LocalDateTime.now(clock)
+    val description = "Licence activated for $licenceId"
+    return HMPPSDomainEvent(
+      eventType,
+      additionalInformation,
+      "https://create-and-vary-a-licence-api.hmpps.service.justice.gov.uk/public/licences/id/$licenceId",
```
andrewrlee, line 33: rather than hard coding this could do with injecting the self link to allow this to work in dev/preprod.
there's an example in `NotifyService`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/OutboundEventsPublisher.kt
```diff
@@ -0,0 +1,45 @@
...
+
+  private val domainEventsTopic by lazy {
+    hmppsQueueService.findByTopicId(DOMAIN_TOPIC_ID)
+      ?: throw NoSuchElementException("Topic with name $DOMAIN_TOPIC_ID doesn't exist")
+  }
+  private val domainEventsTopicClient by lazy { domainEventsTopic.snsClient }
+
+  fun publishDomainEvent(event: HMPPSDomainEvent, licenceId: String) {
+    log.debug("Event {} for id {}", event.eventType, licenceId)
+    val eventType = event.eventType
```
andrewrlee, line 31: super minor, would suggest either referring to `event.eventType` everywhere or pushing it up a line and referring to it in the debug log line and when building up the message attributes

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/DeactivateLicencesService.kt
```diff
@@ -71,6 +74,13 @@ class DeactivateLicencesService(
           eventDescription = "${reason ?: "Licence deactivated automatically as it passed release date"} for ${licence.forename} ${licence.surname}",
         ),
       )
+
+      domainEventsService.recordDomainEvent(
+        LicenceDomainEventType.LICENCE_INACTIVATED,
```
andrewrlee, line 79: if moving the event type on to kind, it would be worth referring to it here: `licence.kind.inactivatedDomainEventType` as you're working over a list of generic licences

(It's unlikely that this would ever be called with variations but would make sense to avoid the hard coding)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceOverrideService.kt
andrewrlee: this could be simplified further by the suggestion below to move to LicenceKind

detekt-baseline.xml
```diff
@@ -2,12 +2,14 @@
 <SmellBaseline>
   <ManuallySuppressedIssues></ManuallySuppressedIssues>
   <CurrentIssues>
+    <ID>CyclomaticComplexMethod:LicenceService.kt$LicenceService$@Transactional fun updateLicenceStatus(licenceId: Long, request: StatusUpdateRequest)</ID>
```
andrewrlee, line 5: by refactoring a bit should hopefully be able to avoid adding to these warnings

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/DeactivateLicencesTest.kt
```diff
@@ -7,18 +7,29 @@ import org.junit.jupiter.api.AfterAll
...
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.domainEvents.DomainEventsService
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.domainEvents.OutboundEventsPublisher
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
 import java.time.Duration
 
 class DeactivateLicencesTest : IntegrationTestBase() {
+  @MockBean
+  private lateinit var eventsPublisher: OutboundEventsPublisher
+
+  private val eventCaptor = argumentCaptor<DomainEventsService.HMPPSDomainEvent>()
```
andrewrlee, line 32: meh.. these have state that won't be reset between tests, would be fine at the moment as you only have one test but it can be a bit confusing if people added extra tests and tried to reuse this so would generally advise not declaring these as fields

The kotlin mockito argument captor stuff is quite nice too 

If you thought there was a good reason to reuse across tests you could use the `@Captor` annotation which should reset it between tests. (Might need to set up some other mockito stuff, not quite sure how it works with junit5)

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/LicenceActivationIntegrationTest.kt
```diff
@@ -33,6 +45,18 @@ class LicenceActivationIntegrationTest : IntegrationTestBase() {
...
+
+    assertThat(eventCaptor.allValues[0].eventType).isEqualTo(DomainEventsService.LicenceDomainEventType.LICENCE_ACTIVATED.value)
+
+    assertThat(eventCaptor.allValues[1].eventType).isEqualTo(DomainEventsService.LicenceDomainEventType.LICENCE_ACTIVATED.value)
+
+    assertThat(eventCaptor.allValues[2].eventType).isEqualTo(DomainEventsService.LicenceDomainEventType.LICENCE_ACTIVATED.value)
+
+    assertThat(eventCaptor.allValues[3].eventType).isEqualTo(DomainEventsService.LicenceDomainEventType.LICENCE_INACTIVATED.value)
+
+    assertThat(eventCaptor.allValues[4].eventType).isEqualTo(DomainEventsService.LicenceDomainEventType.LICENCE_INACTIVATED.value)
```
andrewrlee, line 58: Could maybe  
```kt
val activationEvents = eventCaptor.allValues.filter { it.eventType ==  LICENCE_ACTIVATED }
assertThat(activationEvents).hasSize(3)

val inactivatedEvents = eventCaptor.allValues.filter { it.eventType ==  LICENCE_INACTIVATED }
assertThat(inactivatedEvents).hasSize(3)
```

or if you think event order matters, map to event status and to 
```kt
assertThat(eventTypeStatus).containsExactly(LICENCE_ACTIVATED, LICENCE_ACTIVATED, LICENCE_ACTIVATED, LICENCE_INACTIVATED, LICENCE_INACTIVATED)
```

andrewrlee (dismissed):
LGTM, have some suggestions but nothing blocking

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/OutboundEventsPublisher.kt
```diff
@@ -0,0 +1,45 @@
...
+  fun publishDomainEvent(event: HMPPSDomainEvent, licenceId: String) {
+    val eventType = event.eventType
+    log.debug("Event {} for Licence ID {}", eventType, licenceId)
+    domainEventsTopicClient.publish(
+      PublishRequest.builder()
+        .topicArn(domainEventsTopic.arn)
+        .message(objectMapper.writeValueAsString(event))
+        .messageAttributes(
+          mapOf(
+            "eventType" to MessageAttributeValue.builder().dataType("String").stringValue(event.eventType).build(),
```
andrewrlee, line 38: meh.. could be `eventType`

gurps-singh: Will change in next PR

PR #443, opened 2024-01-24
CVSL-1619 implementing hardstop flag

andrewrlee:
Also renamed a few functions

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
@@ -11,6 +11,7 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.AuditEvent
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.CommunityOffenderManager
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.CrdLicence
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.HardStopLicence
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.PrisonCaseAdministrator
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.VariationLicence
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.Licence
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.LicenceSummary
@@ -317,7 +318,7 @@ class LicenceService(
     val updatedLicence = when (licenceEntity) {
       is CrdLicence -> licenceEntity.submit(submitter as CommunityOffenderManager)
       is VariationLicence -> licenceEntity.submit(submitter as CommunityOffenderManager)
-      is HardStopLicence -> TODO("Submitting hard stop licences not supported yet")
+      is HardStopLicence -> licenceEntity.submit(submitter as PrisonCaseAdministrator)
       else -> error("Unexpected licence type: $licenceEntity")
     }
 
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
@@ -32,6 +32,7 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.CrdLicence
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.Licence
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.LicenceEvent
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.OmuContact
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.PrisonCaseAdministrator
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.VariationLicence
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.LicenceSummary
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.StatusUpdateRequest
@@ -817,7 +818,7 @@ class LicenceServiceTest {
   }
 
   @Test
-  fun `submit a licence saves new fields to the licence`() {
+  fun `submit a CRD licence saves new fields to the licence`() {
     val expectedCom = CommunityOffenderManager(
       staffIdentifier = 2000,
       username = "smills",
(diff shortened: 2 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
```diff
@@ -139,10 +137,6 @@ abstract class Licence(
     return false
   }
 
-  fun isInHardStopPeriod(): Boolean {
-    return IS_IN_HARD_STOP_PERIOD
-  }
-
```
andrewrlee, line 145: the logic of this is calculated in a logic so moved it into the service layer

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/DatesController.kt
```diff
@@ -99,6 +99,6 @@ class DatesController(
     ],
   )
   fun currentHardStopCutoffDate(): HardStopCutoffDate {
-    return HardStopCutoffDate(releaseDateService.getCutOffDateForLicenceTimeOut(LocalDate.now()))
```
andrewrlee, line 102: we have a clock internal to the service so using this but support overriding for tests

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/jobs/ActivateLicencesController.kt
```diff
@@ -48,6 +48,6 @@ class ActivateLicencesController(
     ],
   )
   fun runLicenceActivationJob() {
-    return licenceActivationService.licenceActivationJob()
```
andrewrlee, line 51: made names a bit consistent

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
```diff
@@ -15,29 +17,40 @@ class ReleaseDateService(
...
-      ::excludeBankHolidaysAndWeekends,
+      ::isBankHolidayOrWeekend,
     )
-    return LocalDate.now(clock).isEqual(dateBeforeXWorkingDays) || LocalDate.now(clock).isAfter(dateBeforeXWorkingDays)
+    return LocalDate.now(clock) >= warningThreshold
   }
 
   /** Friday is not considered as weekend */
-  fun excludeBankHolidaysAndWeekends(releaseDate: LocalDate?): Boolean {
+  fun isBankHolidayOrWeekend(releaseDate: LocalDate?): Boolean {
```
andrewrlee, line 53: methods returning booleans generally should be named: `is*`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
```diff
@@ -15,29 +17,40 @@ class ReleaseDateService(
...
+  fun isInHardStopPeriod(licence: Licence, now: Clock? = null): Boolean {
+    val timedOutDate = getCutOffDateForLicenceTimeOut(now)
+    val releaseDate = licence.actualReleaseDate ?: licence.conditionalReleaseDate
+
+    if (releaseDate == null) {
+      log.warn("Licence with id: ${licence.id} has no CRD or ARD")
+      return false
+    }
+    return releaseDate <= timedOutDate
+  }
```
andrewrlee, line 36: new function!

andrewrlee, line 36: new function!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
andrewrlee, line 53: methods returning booleans generally should be named: `is*`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -164,7 +148,7 @@ fun toHardstop(
   licenceVersion = licence.licenceVersion,
   earliestReleaseDate = earliestReleaseDate,
   isEligibleForEarlyRelease = isEligibleForEarlyRelease,
-  isInHardStopPeriod = licence.isInHardStopPeriod(),
+  isInHardStopPeriod = isInHardStopPeriod,
```
andrewrlee, line 151: not super convinced we need this? wouldn't this always be true?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/IntegrationTestBase.kt
```diff
@@ -72,8 +71,6 @@ abstract class IntegrationTestBase {
   @SpyBean
   protected lateinit var hmppsSqsPropertiesSpy: HmppsSqsProperties
 
-  internal val gson = GsonBuilder().setPrettyPrinting().create()
```
andrewrlee, line 75: not used

louisenorris (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/jobs/ActivateLicencesController.kt
SebNegahbanMoJ: Minor and non-blocking, should this be `activateLicences()` to match `deactivateLicences()`?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
SebNegahbanMoJ: Mentioned in Slack too but also putting here for visibility, would we not want it to be `false` after release, at which point we can't default to `true` as there will be a period between release and PP review?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
SebNegahbanMoJ: Forgive me if I'm misunderstanding, is this not saying that it will return true if the licence release date is before or equal to the hard-stop window start? ie it would be true from licence creation up to 1 day before release?

If I'm understanding correctly, I believe it should be the inverse:
`return timedOutDate <= releaseDate && releaseDate >= now`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/jobs/ActivateLicencesController.kt
andrewrlee: it's a bit odd though as this does more than activate licences - it deactivates licences too! 

I called it licenceActivation to represent some more abstract process but didn't think about it too hard as its possibly doing too much

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
SebNegahbanMoJ: Very minor, non-blocking, and if others disagree please ignore

Could we rename `timedOutDate` to `endOfHardStopPeriod` or similar?

SebNegahbanMoJ (approved):

PR #444, opened 2024-01-25
CVSL-1619 Do not consider to be in hard stop period after released

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
```diff
@@ -24,15 +24,20 @@ class ReleaseDateService(
...
     if (releaseDate == null) {
       log.warn("Licence with id: ${licence.id} has no CRD or ARD")
       return false
     }
-    return releaseDate <= timedOutDate
+    return if (releaseDate < LocalDate.now(now)) {
+      false
+    } else {
+      releaseDate <= endOfHardStopPeriod
+    }
```
SebNegahbanMoJ, line 40: Functionally fine, but would
`return releaseDate >= LocalDate.now(now) && releaseDate <= endOfHardStopPeriod` be more succinct?

nishanthmahasamudram (approved):
LGTM

PR #448, opened 2024-02-02
Cvsl 1626 cvl sar filter returned data

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Views.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+class Views {
+  interface PublicSar
```
andrewrlee, line 4: can we just call the `SubjectAccessRequest` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
```diff
@@ -37,24 +38,29 @@ sealed interface Licence {
   val id: Long
```
andrewrlee, line 38: if we're exposing licence id on events we should probably expose it here too

andrewrlee: should probably expose kind too

nishanthmahasamudram: we are already exposing kind.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalCondition.kt
```diff
@@ -8,29 +9,43 @@ data class AdditionalCondition(
...
 
   @Schema(description = "Sequence of this additional condition within the additional conditions", example = "1")
   val sequence: Int? = null,
 
-  @Schema(description = "The textual value for this additional condition", example = "You must not enter the location [DESCRIPTION]")
+  @Schema(
+    description = "The textual value for this additional condition",
+    example = "You must not enter the location [DESCRIPTION]",
+  )
+  @get:JsonView(Views.PublicSar::class)
```
andrewrlee, line 30: think we're ok without this - as it should be covered in the expanded text

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/SubjectAccessRequestController.kt
```diff
@@ -27,6 +29,7 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.publicApi.S
 @RequestMapping("/public", produces = [MediaType.APPLICATION_JSON_VALUE])
```
andrewrlee, line 29: Just noticed this, but it shouldnt be on the /public path

src/test/resources/test_data/sar_content/hardstopLicence.json
```diff
@@ -0,0 +1,89 @@
...
+            "code": "notBreakLaw",
+            "text": "Do not break any law"
+          },
+          {
+            "code": "attendMeetings",
+            "text": "Attend meetings"
+          }
+        ],
+        "standardPssConditions": [],
+        "additionalLicenceConditions": [],
```
andrewrlee, line 41: shouldn't this have an additional condition fed through? its added by default through test data

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/SubjectAccessRequestController.kt
nishanthmahasamudram: private?

src/test/resources/test_data/sar_content/hardstopLicence.json
nishanthmahasamudram: No I am passing emptyMap() for conditionSubmissionStatus

nishanthmahasamudram: Updated the test case to fetch additional conditions

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/TestData.kt
```diff
@@ -28,7 +28,7 @@ object TestData {
     licence = licence,
     id = 2L,
     conditionSequence = 1,
-    conditionType = HARD_STOP_CONDITION.type,
+    conditionType = "AP",
```
andrewrlee, line 31: good spot!

PR #450, opened 2024-02-08
CVSL-1483 change check to run all eligibility checks against a person in prison

gurps-singh:
This PR follows testing of CVSL-1483 where we have found that just having two existing licence checks for an ARD/CRD in the future and an active prison status is problematic, when checking existing licences. Instead, for any person in prison with a non probation licence status :

- `IN_PROGRESS`
- `SUBMITTED`
- `APPROVED`

we will carry out a full eligibility check and will continue to do so while their licence is in any of the statuses above. If the person has a licence in the status above and fails eligibility, they will not be returned in the results as they are now ineligible to use CVL. We would only stop checking eligibility when their licence becomes `ACTIVE` or another probation licence status (`VARIATION_*`)


louisenorris (approved):
LGTM

SebNegahbanMoJ (approved):

PR #451, opened 2024-02-08
CVSL-1517 amend detail url

gurps-singh:
This PR is following feedback from the Delius integration team that the `detailUrl` is pointing to the frontend and not the API so this means the link is obsolete. Changed the link to ensure the public API licence URL points to the API and not the frontend.


louisenorris (approved):
LGTM

PR #453, opened 2024-02-13
populating responsible com full name

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -185,6 +187,8 @@ fun toVariation(
   comUsername = licence.responsibleCom!!.username,
   comStaffId = licence.responsibleCom!!.staffIdentifier,
   comEmail = licence.responsibleCom!!.email,
+  responsibleComFullName = with(licence.responsibleCom!!) { "$firstName $lastName" },
+  updatedByFullName = "",
```
nishanthmahasamudram, line 191: updatedByFullName has to be implemented as part of a new ticket.
This is just a placeholder for now.

andrewrlee (approved):
LGTM

PR #455, opened 2024-02-14
CVSL-1671 fix error returning invalid licence statuses

gurps-singh:
This PR is to allow the Public API to return licences of all statuses when using the `getLicenceById` endpoint on the Public API. This is in response to feedback from the Delius integration team that domain events where the licence has been inactivated can no longer be viewed when the `detailUrl` part of the event is used.


andrewrlee (approved):
LGTM

PR #456, opened 2024-02-15
CVSL-1677 add application event publisher

gurps-singh:
This PR is in response to the Delius integrations team's question around ensuring that we raise events once the transaction for activation and inactivations are committed to the DB. It replaces the `outboundEventsPublisher` calls to publish with its own call to `publishEvent` which then searches for the relevant process to action based on the `@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)` annotation.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/OutboundEventsPublisher.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/OutboundEventsPublisher.kt
@@ -8,8 +8,11 @@ import org.springframework.transaction.event.TransactionPhase
 import org.springframework.transaction.event.TransactionalEventListener
 import software.amazon.awssdk.services.sns.model.MessageAttributeValue
 import software.amazon.awssdk.services.sns.model.PublishRequest
+import software.amazon.awssdk.services.sns.model.PublishResponse
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.domainEvents.DomainEventsService.HMPPSDomainEvent
 import uk.gov.justice.hmpps.sqs.HmppsQueueService
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.TimeUnit
 
 @Service
 class OutboundEventsPublisher(
@@ -19,6 +22,7 @@ class OutboundEventsPublisher(
 
   companion object {
     const val DOMAIN_TOPIC_ID = "domainevents"
+    const val FUTURE_TIMEOUT = 10000L
     private val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/OutboundEventsPublisher.kt
```diff
@@ -26,9 +28,10 @@ class OutboundEventsPublisher(
   }
   private val domainEventsTopicClient by lazy { domainEventsTopic.snsClient }
 
-  fun publishDomainEvent(event: HMPPSDomainEvent, licenceId: String) {
+  @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
```
gurps-singh, line 31: Added the event listener with an `AFTER_COMMIT` phase which, as it suggests, handles the event after the commit has completed successfully

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/OutboundEventsPublisher.kt
```diff
@@ -26,9 +28,10 @@ class OutboundEventsPublisher(
...
+  fun publishDomainEvent(event: HMPPSDomainEvent) {
     val eventType = event.eventType
-    log.debug("Event {} for licence ID {}", eventType, licenceId)
+    log.debug("Event {} for licence ID {}", eventType, event.additionalInformation?.licenceId)
```
andrewrlee, line 34: We're generating duplicate logs in app insights for this - 2 for each event raised as we're not filtering out `debug`.. maybe it would be worth consolidating these messages into one and log it before the event is raised? 

The log line in the `also` doesn't really have enough contextual info
There should also be a `dependency` that would confirm the event is raised if investigating an issue

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/OutboundEventsPublisher.kt
```diff
@@ -26,20 +32,28 @@ class OutboundEventsPublisher(
...
   private val domainEventsTopicClient by lazy { domainEventsTopic.snsClient }
 
-  fun publishDomainEvent(event: HMPPSDomainEvent, licenceId: String) {
+  @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
+  fun publishDomainEvent(event: HMPPSDomainEvent) {
     val eventType = event.eventType
-    log.debug("Event {} for licence ID {}", eventType, licenceId)
-    domainEventsTopicClient.publish(
+    val publishedDomainEvent = buildDomainEvent(event).get(FUTURE_TIMEOUT, TimeUnit.MILLISECONDS)
+    if (publishedDomainEvent != null) {
```
andrewrlee, line 39: Rather than `buildDomainEvent` this is raising the event.  `publishedDomainEvent` is really the `publishResponse`

I don't think it can ever be null. 
* If the future hasn't completed then a timeout exception will be thrown
* If the future completes with an exception, then an ExecutionException will be thrown with the exception as its cause
* If the future completes successfully the object returned will be the response.

andrewrlee (approved):
LGTM

PR #461, opened 2024-02-23
CVSL-1517 add integration test to test whether the event does publish to the topic

gurps-singh:
This PR is to add in an integration test scenario to ensure we are testing that an event does actually publish to the topic successfully and that we can ping the queue to see the event on the queue.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/domainEvents/OutboundEventsPublisherIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/domainEvents/OutboundEventsPublisherIntegrationTest.kt
@@ -1,0 +1,54 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.integration.domainEvents
+
+import org.assertj.core.api.Assertions.assertThat
+import org.awaitility.kotlin.await
+import org.awaitility.kotlin.matches
+import org.awaitility.kotlin.untilCallTo
+import org.junit.jupiter.api.Test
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/domainEvents/OutboundEventsPublisherIntegrationTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        .message(mapper.writeValueAsString(sentEvent))
+        .messageAttributes(
+          mapOf(
+            "eventType" to MessageAttributeValue.builder().dataType("String").stringValue(sentEvent.eventType).build(),
+          ),
+        )
+        .build(),
+    ).get()
+
+    await untilCallTo { domainEventsSqsClient.countMessagesOnQueue(domainEventsQueueUrl).get() } matches { it != 0 }
```
andrewrlee, line 46: I think I found that sometimes the publish happened and the client received the message before this line could be called - if we notice this failing intermittently that might be what is happening and we might need to try rewriting it

andrewrlee (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/domainEvents/OutboundEventsPublisherIntegrationTest.kt
gurps-singh: Will keep an eye on it - otherwise can rewrite with a `awaitAtMost` like you have added for the HDC API.

PR #463, opened 2024-02-26
CVSL-1719 dependency bump

louisenorris:


build.gradle.kts
```diff
@@ -52,7 +52,7 @@ dependencies {
   implementation("uk.gov.service.notify:notifications-java-client:5.0.0-RELEASE")
 
   // PDF Box - for processing MapMaker file upload to get image / text for exclusion zone
-  implementation("org.apache.pdfbox:pdfbox:3.0.1")
+  implementation("org.apache.pdfbox:pdfbox:2.0.29")
```
louisenorris, line 55: Update to version 3.0.1 required reworking of the ExclusionZoneService as the load method is no longer resolved with this version. Leaving for now to see if the bump as stands is enough to address security alerts.

andrewrlee (approved):
LGTM

PR #464, opened 2024-02-28
CVSL-1725 Extracting out condition routes into their own controller

andrewrlee:
To make space for another licences endpoint


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/ConditionController.kt
```diff
@@ -0,0 +1,427 @@
...
+  fun updateAdditionalConditions(
+    @PathVariable("licenceId") licenceId: Long,
+    @Valid @RequestBody
+    request: AdditionalConditionsRequest,
+  ) {
+    return licenceConditionService.updateAdditionalConditions(licenceId, request)
+  }
+
+  /**
+   * This functionality can be replaced once proper licence versioning is implemented
```
SebNegahbanMoJ, line 293: Not a note for this PR or work to do now, but now that we have versioning, presumably we should review whatever this was referring to? 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/ConditionController.kt
```diff
@@ -0,0 +1,427 @@
...
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+    ],
+  )
+  fun addAdditionalCondition(
+    @PathVariable licenceId: Long,
+    @PathVariable conditionType: String,
```
SebNegahbanMoJ, line 163: Not something to resolve now as I appreciate it's a lift-and-shift, but do we actually need this? Or is it just for url mapping?

SebNegahbanMoJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/ConditionController.kt
andrewrlee: I think ideally we'd change the url but its a bit faffy

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/ConditionController.kt
andrewrlee: I think this refers to us not setting the standard conditions via the frontend - we should be able to set/refresh them at various points in the licence life cycle purely from the backend

andrewrlee: I'll update the comment - not that it helps much but at least adds the context!

PR #469, opened 2024-02-29
CVSL-1710 add migration to set updated_by_id

gurps-singh:
This PR introduces the `updated_by_id` column which we need in order to show the full name of the person who last updated the licence on the CA screen. This includes the modelling of the column to set the `updatedByFullName` field correctly as well as a migration to populate the value where possible for existing data.


gurps-singh:
Two outstanding questions (which could be a subsequent PR):
- Are we changing to using the staff member everywhere `updatedByUsername` is used (i.e should we be updating the `updatedByUsername` and setting `updatedBy` everywhere we currently have `updatedByUsername` - eg in `submit`, `timeout`, `deactivate` ... I think yes we should to ensure consistent data around who updated what
- What about licence and audit events which use the username - should these be set using the `staffMember?.username ?: SYSTEM_USER` as well?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
```diff
@@ -128,8 +128,16 @@ abstract class Licence(
   @ManyToOne(fetch = FetchType.EAGER)
   @JoinColumn(name = "responsible_com_id", nullable = false)
   var responsibleCom: CommunityOffenderManager? = null,
+
+  @ManyToOne(fetch = FetchType.EAGER)
+  @JoinColumn(name = "updated_by_id", nullable = true)
+  var updatedBy: Staff? = null,
```
SebNegahbanMoJ, line 134: Do we need to do performance testing on this or is it assumed to be OK as it's the same pattern as we used for `responsibleCom`?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
```diff
@@ -148,46 +156,50 @@ abstract class Licence(
...
-    updatedByUsername: String?,
+    staffMember: Staff?,
   ): Licence
 
-  fun updateAppointmentAddress(appointmentAddress: String?, updatedByUsername: String?) {
+  fun updateAppointmentAddress(appointmentAddress: String?, staffMember: Staff?) {
     this.appointmentAddress = appointmentAddress
     this.dateLastUpdated = LocalDateTime.now()
-    this.updatedByUsername = updatedByUsername
+    this.updatedByUsername = staffMember?.username ?: SYSTEM_USER
```
SebNegahbanMoJ, line 165: I'm not sure if we need to default to `SYSTEM_USER` here, as I don't think there's a scenario where the system can update this info.
The only time I can see it being set to `SYSTEM_USER` would be if for some reason the `staffMember` doesn't exist, in which case we should probably know

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/VariationLicence.kt
```diff
@@ -303,37 +308,40 @@ class VariationLicence(
...
     supersededDate = supersededDate,
     submittedDate = submittedDate,
     licenceActivatedDate = licenceActivatedDate,
+    updatedBy = staffMember ?: this.updatedBy,
   )
 
   override fun overrideStatus(
     statusCode: LicenceStatus,
-    updatedByUsername: String?,
+    staffMember: Staff?,
```
SebNegahbanMoJ, line 323: I don't think this will ever be a staffMember, as it's our support team that do these overrides. Unless they are also present in the staff table, in which case the opposite is true and it should always be populated and shouldn't default IMO

SebNegahbanMoJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
gurps-singh: Tested this locally and everything set correctly. Had a chat with Louise re testing the migration. Will test in dev but it is preprod, where we can see how performant everything is.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
gurps-singh: Had a chat with Andy about this and will seek clarification, we are working on these rules around where username is used and the `updated_by_id` column needs to be populated:
```
If you can't find a user, set the username to SYSTEM_USER
If you have a user, set the username to their username

If you can't find a user, leave the updated_by_id as it was and do not update
If you have a user, set the updated_by_id to the id of the user
```

Happy to jump on a call about this.

gurps-singh: Confirmed, logic should apply in all places we are setting username

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/VariationLicence.kt
gurps-singh: Confirmed, logic should apply in all places we are setting username

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
andrewrlee: I think if we generally support those rules that should be fine - allow providing an optional updater rather than mandating that there is one.

Probably simpler to be consistent - at the small risk of the possibility of bugs if the username is not passed through.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/VariationLicence.kt
andrewrlee: I think all our staff are nomis users so will be recorded in the staff table.

PR #470, opened 2024-03-06
CVSL-1710 add migration to set updated_by_id

gurps-singh:
This PR introduces the `updated_by_id` column which we need in order to show the full name of the person who last updated the licence on the CA screen. This includes the modelling of the column to set the `updatedByFullName` field correctly as well as a migration to populate the value where possible for existing data.

Old PR (got into a rebase nightmare) - https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/469

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
@@ -1,17 +1,16 @@
 package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
 
 import jakarta.persistence.EntityNotFoundException
-import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.AdditionalCondition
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
```diff
@@ -289,24 +289,4 @@ class AuditService(
       changes = changes,
```
andrewrlee, line 289: is there an optimisation we can do here - now we're pretty much always attempting to load up the staff member could we change this to take a nullable staff record rather than doing the look up again here? 

(the places where we're not passing a staff record would then be a bit more obvious as we'd haver to explicitly pass null through) 

If that seems like something worth doing we could do it as a separate follow up PR? (but wouldn't necessarily want this going live/being tested with out it)

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
gurps-singh: Good spot Andy. Yup that would definitely help - just making the changes and will add to this PR

andrewrlee (approved):
LGTM

PR #473, opened 2024-03-13
CVSL-1710 add submittedByFullName to all licence and summary models

gurps-singh:
This PR is to add `submittedByFullName` to the CRD, Hard Stop and Variation licence models.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
@@ -76,6 +76,8 @@ fun transformToLicenceSummary(licence: EntityLicence): LicenceSummary {
       is HardStopLicence -> licence.reviewDate == null
       else -> false
     },
+    updatedByFullName = licence.getUpdatedByFullName(),
+    submittedByFullName = licence.getSubmittedByFullName(),
   )
 }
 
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -464,3 +467,26 @@ fun CaseloadResult.transformToUnstartedRecord(
...
+  val staffMember = when (this) {
+    is HardStopLicence -> this.submittedBy
+    is CrdLicence -> this.submittedBy
+    is VariationLicence -> this.submittedBy
+    else -> null
+  }
+  return if (staffMember != null) {
+    "${staffMember.firstName} ${staffMember.lastName}"
+  } else {
+    ""
```
andrewrlee, line 490: might be better to return null if there was no one.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -464,3 +467,26 @@ fun CaseloadResult.transformToUnstartedRecord(
...
+  }
+}
+
+fun Licence.getSubmittedByFullName(): String {
+  val staffMember = when (this) {
+    is HardStopLicence -> this.submittedBy
+    is CrdLicence -> this.submittedBy
+    is VariationLicence -> this.submittedBy
+    else -> null
```
andrewrlee, line 485: maybe better `error("some reason")` here, so we fail loudly.

andrewrlee (approved):
LGTM

PR #475, opened 2024-03-14
CVSL-1741 add endpoint to get licences for approval

gurps-singh:
This PR is to introduce a new endpoint to retrieve the licences which need approving for the CA. This replaces the call in the frontend for `getLicencesForApproval`.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ApprovedLicenceSummary.kt
```diff
@@ -0,0 +1,59 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceKind
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceType
+import java.time.LocalDateTime
+
```
gurps-singh, line 9: This currently matches what we have as a licence in `ManagedCase` in `server/@types/managedCase.ts` in the frontend with the addition of the CRN. Not included any release date (conditional release date) in particular which I am still a bit on the fence on as the results are sorted by the conditional release date but we are not sending that back as part of the response?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/request/ApproveLicencesRequest.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.request
+
+import io.swagger.v3.oas.annotations.media.Schema
+
```
gurps-singh, line 4: Looking at the `getLicencesForApproval` call on the frontend, the only thing we require are the `filteredPrisons` as the other values are either null or can be plugged straight into the query to retrieve the relevant licences (see `LicenceRepository` and `LicenceService` changes).

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ApprovedLicenceSummary.kt
andrewrlee: there's a risk that there were only enough things in that licence type that were required for compilation and that the nunjucks template referred to more. 

The type isn't very good - should really expand or refer to existing types

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/request/ApproveLicencesRequest.kt
andrewrlee: Yes, that makes sense

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ApprovedLicenceSummary.kt
gurps-singh: As endpoint is not currently being used, will add changes to separate PR.

andrewrlee (approved):
LGTM

PR #477, opened 2024-03-21
CVSL-1724 add working days service as needed to calculate early release cases

gurps-singh:
This PR is to add a WorkingDaysService in the backend to aid calculation of dates when we need to calculate using working days (as needed to change the logic in the `isInHardStopPeriod` in CVSL-1724)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/WorkingDaysService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/WorkingDaysService.kt
@@ -1,0 +1,35 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
+
+import org.springframework.stereotype.Service
+import java.time.DayOfWeek
+import java.time.LocalDate
+
+@Service
+class WorkingDaysService(private val bankHolidayService: BankHolidayService) {
+
+  fun addWorkingDays(date: LocalDate, workingDays: Int): LocalDate {
+    val bankHolidays = bankHolidayService.getBankHolidaysForEnglandAndWales()
+    var adjustedDate = date
+    for (i in 0 until workingDays) {
+      adjustedDate = getNextWorkingDay(bankHolidays, adjustedDate)
+    }
+    return adjustedDate
+  }
+
+  fun isWeekend(date: LocalDate): Boolean {
+    val weekend = listOf(DayOfWeek.SATURDAY, DayOfWeek.SUNDAY)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/WorkingDaysService.kt
```diff
@@ -0,0 +1,35 @@
...
+import java.time.DayOfWeek
+import java.time.LocalDate
+
+@Service
+class WorkingDaysService(private val bankHolidayService: BankHolidayService) {
+
+  fun addWorkingDays(date: LocalDate, workingDays: Int): LocalDate {
+    val bankHolidays = bankHolidayService.getBankHolidaysForEnglandAndWales()
+    var adjustedDate = date
+    for (i in 0 until workingDays) {
```
andrewrlee, line 13: you can probably remove this unused `i` variable somehow.

We calculate something very similar to this in the ReleaseDateService.

andrewrlee: I'll merge this now but we could look at merging the two implementations

andrewrlee (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/WorkingDaysServiceTest.kt
```diff
@@ -0,0 +1,125 @@
...
+  fun `is date a non working day`() {
+    val today = LocalDate.of(2024, 3, 23)
+    val isNonWorkingDay = service.isNonWorkingDay(someBankHolidays, today)
+    assertTrue(isNonWorkingDay)
+  }
+
+  @Test
+  fun `is date a working day`() {
+    val today = LocalDate.of(2024, 3, 21)
+    val isWorkingDay = service.isNonWorkingDay(someBankHolidays, today)
```
andrewrlee, line 47: meh.. this should probably be called `isNonWorkingDay`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/WorkingDaysService.kt
gurps-singh: Removed the `I` in place of a repeat. Re release date service, I'll refactor this in next PR to use this instead of the implementations we have in the `ReleaseDateService`

PR #481, opened 2024-03-27
CVSL-1775 updating dependencies

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ExclusionZoneService.kt
```diff
@@ -173,7 +175,7 @@ class ExclusionZoneService(
   fun extractDescription(fileStream: InputStream): String? {
     var pdfDoc: PDDocument? = null
     try {
-      pdfDoc = PDDocument.load(fileStream)
+      pdfDoc = Loader.loadPDF(RandomAccessReadBuffer(fileStream))
```
andrewrlee, line 178: major pdf box dependency change has new api - will need to check pdf functionality still works as expected

louisenorris (approved):
LGTM

PR #486, opened 2024-04-05
CVSL-1786 adding hard stop dates to licence summaries

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceCreationResponse.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "A reference to the created licence")
+data class LicenceCreationResponse(
+  @Schema(description = "Internal identifier for this licence generated within this service", example = "123344")
+  val licenceId: Long,
+)
```
andrewrlee, line 9: we have no need to return full licence summaries on creation so now only returning a summary object with the licence id in it. 
This means that we don't have to hook up release date service to licence creation and simplifies things!

nishanthmahasamudram (approved):
LGTM

PR #487, opened 2024-04-09
CVSL-1723 change functionality of time out job to exclude early releases

gurps-singh:
This PR is to exclude early release candidates from being timed out using the time out job. We now cast a wider net to capture licences and then check whether each of the licences is in the hard stop period, and if they are, we retain the licence to time it out.


andrewrlee (approved):
LGTM

PR #488, opened 2024-04-10
CVSL-1786 Adding hardstop dates to search results

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CaseloadService.kt
```diff
@@ -32,10 +32,4 @@ class CaseloadService(
       ),
     )
   }
-
-  fun PrisonerSearchPrisoner.toSentenceDateHolder() = object : SentenceDateHolder {
```
andrewrlee, line 36: Extracted this out to reuse across search and here

gurps-singh (approved):

PR #490, opened 2024-04-11
CVSL-1786 Fixing date format of cvsl dates

andrewrlee:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/CaseloadIntegrationTest.kt
```diff
@@ -86,6 +86,22 @@ class CaseloadIntegrationTest : IntegrationTestBase() {
...
+      webTestClient.get()
+        .uri(GET_PRISONER)
+        .accept(APPLICATION_JSON)
+        .headers(setAuthorisation(roles = listOf("ROLE_CVL_ADMIN")))
+        .exchange()
+        .expectStatus().isEqualTo(OK.value())
+        .expectHeader().contentType(APPLICATION_JSON)
+        .expectBody()
+        .jsonPath("$.cvl.hardStopDate").value<String> { assertThat(it).matches("\\d{2}/\\d{2}/\\d{4}") }
+        .jsonPath("$.prisoner.dateOfBirth").value<String> { assertThat(it).matches("\\d{4}-\\d{2}-\\d{2}") }
```
andrewrlee, line 103: bit janky that we have two different date formats but didn't want to start moving proxied prisoner offender search methods over to use CVL formatting as that would be loads of frontend changes everywhere.

Kept the CVL fields to follow standard cvl date format... we could really do with just moving everything over!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/CaseloadItem.kt
```diff
@@ -13,9 +14,11 @@ data class CvlFields(
   val licenceType: LicenceType,
 
   @Schema(description = "Date which the hard stop period will start", example = "2023-05-01")
+  @JsonFormat(pattern = "dd/MM/yyyy")
   var hardStopDate: LocalDate? = null,
 
   @Schema(description = "Date which to show the hard stop warning", example = "2023-05-01")
+  @JsonFormat(pattern = "dd/MM/yyyy")
```
louisenorris, line 21: Should the example match the pattern format?

louisenorris (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/CaseloadItem.kt
andrewrlee: yes, it should! - thanks will fix in follow up

PR #493, opened 2024-04-11
CVSL-1727 Adding is due for early release flag

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
```diff
@@ -25,7 +25,7 @@ class ReleaseDateService(
     val hardStopDate = getHardStopDate(sentenceDateHolder)
     val today = LocalDate.now(now)
 
-    if (hardStopDate == null) {
+    if (hardStopDate == null || sentenceDateHolder.licenceStartDate == null) {
```
andrewrlee, line 28: some test data failed on this - should check but imagine we probably have some licences without a start date in the DB

gurps-singh (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
SebNegahbanMoJ: We definitely will because of IS91 sentence deactivation etc

PR #501, opened 2024-04-17
Moving to use hard stop in test1

andrewrlee:
Always deploy HEAD of main to test1 and disable test1 branch deploy


nishanthmahasamudram (approved):
LGTM

PR #502, opened 2024-04-17
CVSL-1454 add hard stop licence approved email notification

gurps-singh:
This PR introduces scenario 2 for the hard stop email notifications which is to notify the PP when the approver approves a standard licence for someone. This will then fire the  "_When the prison generate a standard licence for someone_" email template in Notify. Alongside this, this PR contains config to allow us to test Notify on local when the config is enabled (set to default to false in `application-dev.yml`


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -220,6 +220,20 @@ class LicenceService(
...
 
+    if (request.status === APPROVED && licenceEntity is HardStopLicence) {
+      val com = licenceEntity.responsibleCom
+      if (com != null) {
+        notifyService.sendHardStopLicenceApprovedEmail(
+          com.email,
+          licenceEntity.forename ?: "unknown",
+          licenceEntity.surname ?: "unknown",
+          licenceEntity.crn,
+          licenceEntity.conditionalReleaseDate ?: licenceEntity.actualReleaseDate,
```
gurps-singh, line 231: Unsure about what the release date should be in this instance? Is it still CRD if set otherwise ARD?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -220,6 +220,20 @@ class LicenceService(
       notifyReApprovalNeeded(licenceEntity)
     }
 
+    if (request.status === APPROVED && licenceEntity is HardStopLicence) {
```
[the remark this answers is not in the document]
andrewrlee: yes think so - also could maybe extract body of if statement in to helper function as this method is horrible as it is

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -220,6 +220,20 @@ class LicenceService(
...
+      val com = licenceEntity.responsibleCom
+      if (com != null) {
+        notifyService.sendHardStopLicenceApprovedEmail(
+          com.email,
+          licenceEntity.forename ?: "unknown",
```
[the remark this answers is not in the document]
andrewrlee: think better to blow up if not set

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
andrewrlee: think this is right, could use licenceStartDate??? @SebNegahbanMoJ what do you think?

SebNegahbanMoJ: If we were going to be explicit it should be ARD ?: CRD, as CRD should always be set and ARD is the `confirmed` date.
Given that's the case though, yeah we can probably use `licenceStartDate`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -301,6 +307,20 @@ class LicenceService(
     )
   }
 
+  private fun notifyHardStopApprovalNeeded(licenceEntity: EntityLicence) {
```
andrewrlee, line 310: meh.. `notifyComAboutHardstopLicenceApproval` ?

andrewrlee (approved):
LGTM

PR #504, opened 2024-04-18
CVSL-1454 add notification for previously approved now edited and timed out licence

gurps-singh:
This PR is the second for the hard stop email notifications. This introduces an email where if the person has a previously approved licence, which is being edited and been timed out, an email notification is sent.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/UpdateSentenceDateService.kt
```diff
@@ -72,6 +72,9 @@ class UpdateSentenceDateService(
         recordAuditEvent(updatedLicenceEntity, "Sentence dates updated")
         recordAuditEvent(timedOutLicence, "Licence automatically timed out after sentence dates update")
         recordLicenceEvent(timedOutLicence, "Licence automatically timed out after sentence dates update")
```
andrewrlee, line 74: wonder if we had a:
```kt
licenceService.timeout(licence, reason = ""Licence automatically timed out after sentence dates update")
```
that we could use in both places - which did the `timeout` the persisting, the auditing and the notifying?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/TimeOutLicencesService.kt
```diff
@@ -76,6 +80,22 @@ class TimeOutLicencesService(
           eventDescription = "${reason ?: "Licence automatically timed out"} for ${licence.forename} ${licence.surname}",
         ),
       )
+
+      // previously approved now edited licence that is timed out
+      if (hardStopEnabled && licence.versionOfId != null) {
```
andrewrlee, line 85: presuambly we don't need to add this flag as it we would only run this job if we were enabling hardstop

andrewrlee: if we were to use it - would make more sense exiting early at the start of the job

andrewrlee (approved):
LGTM

PR #506, opened 2024-04-18
CVSL-1824 Fix potentially null fields

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison/PrisonerSearchPrisoner.kt
```diff
@@ -10,7 +10,7 @@ data class SearchPaginationResponse(
 data class PrisonerSearchPrisoner(
   val prisonerNumber: String,
   val pncNumber: String? = null,
-  val bookingId: String,
+  val bookingId: String? = null,
```
andrewrlee, line 13: the main issues are in this file

PR #509, opened 2024-04-19
CVSL-1454 add notification for hard stop licences not reviewed emails

gurps-singh:
This PR is the final scenario for the hard stop email notifications. This sends a notification when a review of a standard licence is overdue. This is 5 days after the licence activated date and where a review has not occurred.


andrewrlee (approved):
LGTM

PR #511, opened 2024-04-22
CVSL-1822 add extra fields for PP search results

gurps-singh:
This PR is to add two additional fields to the search response for the PP search.


nishanthmahasamudram (approved):
LGTM

PR #516, opened 2024-04-25
CVSL-1860 Record event when standard licence is reviewed with no variation

andrewrlee:
This is not used but probably better to capture it now. 
Also differentiating between review types - again better to do this up front rather than having to fix without the data


nishanthmahasamudram (approved):
LGTM

PR #518, opened 2024-04-26
CVSL-1829 Adding isEligibleForEarlyRelease to caseload item

andrewrlee:
This is being recalculated on the frontend when there is no licence - this means we can remove that bit of logic


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
```diff
@@ -62,6 +66,11 @@ class ReleaseDateService(
     return LocalDate.now(clock) >= warningThreshold
   }
 
+  fun isEligibleForEarlyRelease(sentenceDateHolder: SentenceDateHolder): Boolean {
```
nishanthmahasamudram, line 69: Can we use same method name with different param types?

nishanthmahasamudram (approved):
LGTM

PR #519, opened 2024-04-26
CVSL-1832 add updatedByFullName to licence summary

gurps-singh:
This PR is to address a bug where Last Worked On By was not populated in the frontend in the CA view.


andrewrlee (approved):
LGTM

PR #522, opened 2024-05-01
CVSL-1771 add post recall release date to DB

gurps-singh:
This PR is to add the new post recall release date field in the DB and on the entities.


docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: "3"
```
gurps-singh, line 1: The script was complaining that this is no longer needed so removed it.

PR #523, opened 2024-05-01
CVSL-1877 Fixing OOM issue due to non-terminating loop

andrewrlee:
Also increasing memory to help with general OOM issues that we've seen due to pulling back loads of data


Dockerfile
```diff
@@ -37,4 +37,4 @@ COPY --from=builder --chown=appuser:appgroup /app/applicationinsights.dev.json /
 
 USER 2000
 
-ENTRYPOINT ["java", "-javaagent:/app/agent.jar", "-jar", "/app/app.jar"]
+ENTRYPOINT ["java", "-XX:+AlwaysActAsServerClassMachine", "-javaagent:/app/agent.jar", "-jar", "/app/app.jar"]
```
andrewrlee, line 40: pulled back from the template project

helm_deploy/create-and-vary-a-licence-api/values.yaml
```diff
@@ -19,7 +19,7 @@ generic-service:
 
   # Environment variables to load into the deployment
   env:
-    JAVA_OPTS: "-Xmx512m"
+    JAVA_OPTS: "-Xmx768m"
```
andrewrlee, line 22: we tend to see OOM issues occasionally due to pulling back large caselists, we've now doubled the namespace limit to 2048 in CP - so have some more breathing room to increase this.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -963,15 +963,17 @@ class LicenceService(
...
-  private fun findOriginalLicenceForVariation(variationLicence: VariationLicence): CrdLicence {
+  private fun findOriginalLicenceForVariation(variationLicence: VariationLicence): EntityLicence {
     var originalLicence = variationLicence
     while (originalLicence.variationOfId != null) {
       val licence = licenceRepository
         .findById(originalLicence.variationOfId!!)
         .orElseThrow { EntityNotFoundException("${originalLicence.variationOfId}") }
       when (licence) {
         is CrdLicence -> return licence
+        is HardStopLicence -> return licence
```
andrewrlee, line 974: this was not handled so would never terminate

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -963,15 +963,17 @@ class LicenceService(
...
         is VariationLicence -> originalLicence = licence
+        else -> error("Unknown licence type in hierarchy: ${licence.javaClass}")
```
andrewrlee, line 976: Loudly fail if other types are introduced (can't use sealed classes here due to using inheritance)

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceServiceTest.kt
```diff
@@ -448,6 +449,61 @@ class LicenceServiceTest {
...
+        anyList(),
+        any<LocalDate>(),
+      ),
+    ).thenReturn(
+      listOf(
+        activeVariationLicence,
+      ),
+    )
+
+    val licenceSummaries = service.findRecentlyApprovedLicences(emptyList())
```
andrewrlee, line 488: this would OOM without the fix

mattops (approved):

PR #526, opened 2024-05-14
CVSL-1899 Tweak hard stop logic

andrewrlee:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/TimeOutLicencesIntegrationTest.kt
```diff
@@ -61,28 +61,30 @@ class TimeOutLicencesIntegrationTest : IntegrationTestBase() {
...
-          tuple(8L, LicenceStatus.IN_PROGRESS)
-          tuple(9L, LicenceStatus.IN_PROGRESS)
-        } else {
-          tuple(1L, LicenceStatus.TIMED_OUT)
-          tuple(2L, LicenceStatus.TIMED_OUT)
-          tuple(4L, LicenceStatus.TIMED_OUT)
-          tuple(7L, LicenceStatus.TIMED_OUT)
-          tuple(8L, LicenceStatus.TIMED_OUT)
-          tuple(9L, LicenceStatus.TIMED_OUT)
-        },
```
andrewrlee, line 84: this isn't valid syntax to build up a list of varargs - as it's inside a block it effectively only returned the final tuple so equated to:
`.contains(tuple(9L, LicenceStatus.TIMED_OUT))`

The missing commas was a clue - I've swapped this over to compare all elements of a list and chosen list based on whether it's a bank holiday or not

andrewrlee: The logic for if its a weekend or  bank holiday looks a bit suspect too

andrewrlee: I've removed the assertions if the tests are run during the weekend/bankholiday

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/TimeOutLicencesIntegrationTest.kt
```diff
@@ -61,28 +61,22 @@ class TimeOutLicencesIntegrationTest : IntegrationTestBase() {
...
+        .extracting<Tuple> {
+          tuple(it.licenceId, it.licenceStatus)
+        }
+        .containsExactlyElementsOf(
+          listOf(
+            tuple(1L, LicenceStatus.TIMED_OUT),
+            tuple(2L, LicenceStatus.TIMED_OUT),
+            tuple(4L, LicenceStatus.TIMED_OUT),
+            tuple(9L, LicenceStatus.TIMED_OUT),
+          ),
```
andrewrlee, line 77: the other two expected items had ARD > CRD which we don't want to time out anymore

nishanthmahasamudram (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
```diff
@@ -81,33 +81,32 @@ class ReleaseDateService(
...
+      else -> calculateHardStopDate(conditionalReleaseDate)
+    }
+  }
 
+  private fun calculateHardStopDate(conditionalReleaseDate: LocalDate): LocalDate {
+    val date = if (conditionalReleaseDate.isNonWorkingDay()) 1.workingDaysBefore(conditionalReleaseDate) else conditionalReleaseDate
     return 2.workingDaysBefore(date)
   }
 
+  private fun isExcludedFromHardstop(actualReleaseDate: LocalDate, conditionalReleaseDate: LocalDate): Boolean {
```
nishanthmahasamudram, line 98: Is this the replacement for this??
**val isNotAnEarlyRelease = actualReleaseDate >= 1.workingDaysBefore(conditionalReleaseDate)
    val isArdTheReleaseDate = actualReleaseDate <= conditionalReleaseDate**

andrewrlee: yes kind of - previously we would have used ARD as the basis of hard stop if these were true.

Now we will not have a hard stop date if these are true, we also now don't do hard stop if ARD is > than CRD

PR #527, opened 2024-05-15
CVSL-1738 add inactivate recall job to backend

gurps-singh:
This PR is to add the job to inactivate active licences and their variations where a PRRD has been entered, meaning the person has been recalled.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/InactivateRecallLicencesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/InactivateRecallLicencesService.kt
@@ -1,0 +1,33 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.jobs
+
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.repository.LicenceRepository
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.LicenceService
+
+@Service
+class InactivateRecallLicencesService(
+  private val licenceRepository: LicenceRepository,
+  private val licenceService: LicenceService,
+) {
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @Transactional
+  fun inactivateLicences() {
+    log.info("Starting runInactivateRecallLicencesJob")
+    val licences = licenceRepository.getActiveAndVariedLicencesWhichAreNowRecalls()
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/InactivateRecallLicencesService.kt
```diff
@@ -0,0 +1,33 @@
...
+    val licences = licenceRepository.getActiveAndVariedLicencesWhichAreNowRecalls()
+    if (licences.isEmpty()) {
+      log.info("No matching licences to inactivate were found")
+      return
+    }
+    licenceService.inactivateLicences(
+      licences,
+      "Licence automatically inactivated as licence has a PRRD and is now a recall",
+    )
+    log.info("${licences.size} licences were automatically inactivated")
```
andrewrlee, line 31: loving the logging!

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/jobs/InactivateRecallLicencesControllerTest.kt
```diff
@@ -0,0 +1,57 @@
...
+class InactivateRecallLicencesControllerTest {
+  @MockBean
+  private lateinit var inactivateRecallLicencesService: InactivateRecallLicencesService
+
+  @Autowired
+  private lateinit var mvc: MockMvc
+
+  @BeforeEach
+  fun reset() {
+    org.mockito.kotlin.reset(inactivateRecallLicencesService)
```
andrewrlee, line 39: meh.. would generally statically assert this too

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/InactivateRecallLicencesServiceTest.kt
```diff
@@ -0,0 +1,126 @@
...
+    SecurityContextHolder.setContext(securityContext)
+
+    org.mockito.kotlin.reset(
+      licenceRepository,
+      licenceService,
+    )
+  }
+
+  @Test
+  fun `should inactivate active licence with a PRRD of today`() {
```
andrewrlee, line 43: I probably would just have 1 happy path test here - there's no need to test PRRD related to today or variations vs active etc.. - all that logic is in the query that you're stubbing the response to anyway. 

It's potentially a bit misleading as it suggests that the unit under test would treat these differently when really it just inactivates everything its collaborator returns - kind of leaks implementation details of another class and could easily be missed if we change the query

andrewrlee (approved):
LGTM

PR #530, opened 2024-05-23
Cvsl 1750 other tab email cvl support

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/NotifyAttentionNeededLicence.kt
```diff
@@ -0,0 +1,32 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDate
+
+@Schema(description = "Describes a licence that needs attention for support")
+data class NotifyAttentionNeededLicence(
```
andrewrlee, line 8: This isn't an actual api representation so might be better to move into the service package - it doesn't need swagger annotations

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/NotifyAttentionNeededLicencesService.kt
```diff
@@ -0,0 +1,54 @@
...
+  private val licenceRepository: LicenceRepository,
+  private val prisonerSearchApiClient: PrisonerSearchApiClient,
+  private val notifyService: NotifyService,
+) {
+
+  @Transactional
+  fun notifyAttentionNeededLicences() {
+    val attentionNeededLicences = licenceRepository.getAttentionNeededLicences()
+    if (attentionNeededLicences.isEmpty()) {
+      return
```
andrewrlee, line 25: can we log a message saying there were no licences, and could we add a log message saying how many licences we emailed about below?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/NotifyAttentionNeededLicencesServiceTest.kt
```diff
@@ -0,0 +1,209 @@
...
+      emailAddress,
+      fileContent.toByteArray(),
+      fileName,
+    )
+  }
+
+  private companion object {
+    val emailAddress = "testemail@probation.gov.uk"
+    val fileName = "attentionNeededLicences_" + LocalDate.now() + ".csv"
+    val aLicenceEntity = TestData.createCrdLicence().copy(
```
andrewrlee, line 101: The benefit of using copy is that you only need to provide values for relevant fields - this is nice as it draws attention to what attributes are important for this functionality. I would suggest trimming it down to the smallest point that the tests still pass

andrewrlee (approved):
LGTM

PR #535, opened 2024-06-06
CVSL-1738 remove inactivate job

gurps-singh:
Now we are dealing with inactivations differently, the inactivate job is no longer needed.


andrewrlee (approved):
LGTM

PR #538, opened 2024-06-10
CVSL-1941 Adding new endpoint to deactivate active licence and any variations

SebNegahbanMoJ:


nishanthmahasamudram (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/request/DeactivateLicenceAndVariationsRequest.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.request
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Request object for deactivating an active licence and its variations")
+data class DeactivateLicenceAndVariationsRequest(
+  @Schema(description = "A key representing the reason for the variation", example = "RESENTENCED")
+  val reason: String? = null,
```
andrewrlee, line 8: the idea would be that you use the enum here - let the framework handle the conversion

andrewrlee (approved):
LGTM

PR #541, opened 2024-06-17
CVSL-1906 add calls to probation search and community api

gurps-singh:
This PR is to add some extra calls to the clients in preparation of moving the approver caseload to the backend.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
@@ -13,7 +13,6 @@ class ApproverCaseloadService(
 
   fun getApprovalNeeded(prisons: List<String>): List<LicenceSummaryApproverView> {
     val filteredPrisons = prisons.filterNot { it == "CADM" }
-    log.info(filteredPrisons.toString())
     val licences = prisonApproverService.getLicencesForApproval(filteredPrisons)
     if (licences.isEmpty()) {
       return emptyList()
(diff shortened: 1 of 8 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadServiceTest.kt
```diff
@@ -0,0 +1,96 @@
...
+
+  private val service = ApproverCaseloadService(prisonApproverService)
+
+  @BeforeEach
+  fun reset() {
+    reset(prisonApproverService)
+  }
+
+  @Test
+  fun `CADM prison code is filtered out`() {
```
andrewrlee, line 28: to test this you'd need to verify the call to prisonApprover caseload?  and check the CADM caseload isn't passed through

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.LicenceSummaryApproverView
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.PrisonApproverService
+
+@Service
+class ApproverCaseloadService(
+  private val prisonApproverService: PrisonApproverService,
```
andrewrlee, line 11: I split out the `PrisonApproverService` recently so that there would be a place for the caseload logic.

andrewrlee: we've had a bit of a proliferation of services.. maybe we could keep them separate and look to merge later

We want to avoid having multiple layers of transformation but to pull back the changes from the frontend by porting it, keeping it as similar as possible I guess it makes sense to keep the two layers

andrewrlee: Think maybe we need a bit of a modelling session in the next few weeks to try to instill some structure in to the backend

andrewrlee (approved):
LGTM

PR #543, opened 2024-06-18
CVSL-1906 add and amend data classes

gurps-singh:
This PR is to add additional data classes to represent the objects we will need in the approver caseload.


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/config/LocalStackContainer.kt
```diff
@@ -34,7 +34,7 @@ object LocalStackContainer {
       withServices(LocalStackContainer.Service.SQS, LocalStackContainer.Service.SNS)
       withEnv("DEFAULT_REGION", "eu-west-2")
       waitingFor(
-        Wait.forLogMessage(".*Running on.*", 1),
+        Wait.forLogMessage(".*Ready.*", 1),
```
gurps-singh, line 37: Integration tests were not running in IntelliJ for some reason - drilled down to this wait and checked other repos using LocalStackContainer and they were waiting for the container to be ready rather than running on a particular port. Changed and now run as expected.

louisenorris (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/ApprovalCase.kt
```diff
@@ -0,0 +1,18 @@
...
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+data class ApprovalCase(
+  val licenceId: Long?,
+  val name: String?,
+  val prisonerNumber: String?,
+  val submittedByFullName: String?,
+  val releaseDate: LocalDate?,
+  val sortDate: LocalDate?,
```
andrewrlee, line 12: don't need to fix now but think you can get away with just returning releaseDate as a LocalDate as they're the same thing.

One is a date on the frontend for convenience for sorting but can probably sort just on the backend anyway?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/ProbationSearchApiClient.kt
```diff
@@ -53,7 +53,7 @@ class ProbationSearchApiClient(@Qualifier("oauthProbationSearchApiClient") val p
   }
 
   fun searchForPeopleByNomsNumber(
-    nomsIds: List<String?>,
+    nomsIds: List<String>,
```
andrewrlee, line 56: sweet

andrewrlee (approved):
LGTM

andrewrlee (changes requested):
ooh, think packages are wrong as these are representations so need to be in model? also need swagger docs

andrewrlee (approved):
LGTM

PR #545, opened 2024-06-19
CVSL-1906 add approver caseload service to back end

gurps-singh:
This PR is to introduce the approver caseload service to the backend. This includes porting the service that is currently in the frontend to the backend.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
@@ -49,7 +49,7 @@ class ApproverCaseloadService(
 
     val caseload = prisonerRecord.map { (nomisRecord, deliusRecord, licenceSummary) ->
       val releaseDate = when (licenceSummary) {
-        null -> selectReleaseDate(nomisRecord) // you would always have a licence summary here as without it you wouldnt get the nomis ID
+        null -> selectReleaseDate(nomisRecord)
         else -> licenceSummary.actualReleaseDate ?: licenceSummary.conditionalReleaseDate
       }
       ApprovalDetails(
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
@@ -3,24 +3,141 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.caseload
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.ApprovalCase
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.LicenceSummaryApproverView
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.ProbationPractitioner
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.PrisonApproverService
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.convertToTitleCase
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonerSearchApiClient
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonerSearchPrisoner
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.CommunityApiClient
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.OffenderDetail
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.ProbationSearchApiClient
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.User
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
+import java.time.LocalDate
+import java.time.LocalDateTime
 
 @Service
 class ApproverCaseloadService(
   private val prisonApproverService: PrisonApproverService,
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadServiceTest.kt
```diff
@@ -24,4 +45,151 @@ class ApproverCaseloadServiceTest {
     service.getApprovalNeeded(aListOfPrisonCodes)
     verify(prisonApproverService, times(1)).getLicencesForApproval(listOf("ABC", "DEF"))
   }
+
+  @Test
+  fun `It builds the approval needed caseload`() {
```
gurps-singh, line 50: Will be adding more tests in another PR but this matches the unit test on the front end currently.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
```diff
@@ -3,24 +3,141 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.caseload
...
+    val deliusNomisIds = deliusRecords.mapNotNull { it.otherIds.nomsNumber }
+    val nomisRecords = prisonerSearchApiClient.searchPrisonersByNomisIds(deliusNomisIds)
+
+    val prisonerRecord: List<Triple<PrisonerSearchPrisoner, OffenderDetail?, LicenceSummaryApproverView?>> =
+      nomisRecords.map {
+        Triple(it, it.getDeliusRecord(deliusRecords), it.getLicenceSummary(licences))
+      }
+
+    val caseload = prisonerRecord.map { (nomisRecord, deliusRecord, licenceSummary) ->
+      val releaseDate = when (licenceSummary) {
```
gurps-singh, line 51: Umming and ahhing about this as although this is in the frontend, a licence summary wouldn't be null at this point as it is used to retrieve the `nomisIds` which drive the searches in NOMIS and Delius. Don't think this is needed but tried to match what is on the frontend.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
```diff
@@ -3,24 +3,141 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.caseload
...
+      return null
+    }
+    ProbationPractitioner(
+      staffCode = activeCom.staffDetail.code,
+      name = "${activeCom.staffDetail.forenames} ${activeCom.staffDetail.surname}".convertToTitleCase(),
+    )
   }
 }
+
+data class ApprovalDetails(
```
andrewrlee, line 127: **suggestion:** could these be private classes? Presumably they are only referenced in this file?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
```diff
@@ -3,24 +3,141 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.caseload
...
+fun PrisonerSearchPrisoner.getDeliusRecord(deliusRecords: List<OffenderDetail>): OffenderDetail? {
+  return deliusRecords.find { it.otherIds.nomsNumber == this.prisonerNumber }
+}
+
+fun PrisonerSearchPrisoner.getLicenceSummary(licences: List<LicenceSummaryApproverView>): LicenceSummaryApproverView? {
+  val licenceSummaries = licences.filter { it.nomisId == this.prisonerNumber }
+  return if (licenceSummaries.size == 1) licenceSummaries.first() else licenceSummaries.find { it.licenceStatus != LicenceStatus.ACTIVE }
+}
+
+fun PrisonerSearchPrisoner.selectReleaseDate(): LocalDate? {
```
andrewrlee, line 103: **suggestion(non-blocking):** we have this logic on prisoner search prisoner, in lots of places now.. 

Would be good to extract this out and share maybe into its own module in `uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/prison`

Can see it in:
* LicenceFactory x 2
* EligibilityService

(There's an example in prisonerSearchService which looks similar where we fallback to releaseDate which probably should be CRD too)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
```diff
@@ -3,24 +3,141 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.caseload
...
   }
+}
 
-  companion object {
-    val log: Logger = LoggerFactory.getLogger(this::class.java)
+fun PrisonerSearchPrisoner.getDeliusRecord(deliusRecords: List<OffenderDetail>): OffenderDetail? {
+  return deliusRecords.find { it.otherIds.nomsNumber == this.prisonerNumber }
+}
+
+fun PrisonerSearchPrisoner.getLicenceSummary(licences: List<LicenceSummaryApproverView>): LicenceSummaryApproverView? {
```
andrewrlee, line 98: **quibble(non-blocking)**:  I'd probably be tempted to swap the target of the extension function round: 

`licenceSummaries.findFor(prisoner)`

The reason why is that licenceSummaries isn't really an attribute of prisoner search prisoner:
```kt
val prisoners = prisonSearchApi.getPrisoners()
val licence = prisoner.getLicenceSummary()
```
makes it look like the licences are stored in the prisonSearchApi which is potentially a bit confusing.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
andrewrlee: yes, totally right - this was something in hindsight I thought I should have removed. I was copying over from the existing implementation. 

I think remove it?

andrewrlee (approved):
LGTM

PR #548, opened 2024-06-26
CVSL-1906 add endpoint for get approval caseload

gurps-singh:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/CaseloadGenerationController.kt
```diff
@@ -0,0 +1,77 @@
...
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.ApprovalCase
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.resource.Tags
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.caseload.ApproverCaseloadService
+
+@Tag(
+  name = Tags.CASELOAD,
+  description = "Returns the relevant caseload",
+)
+@RestController
+@RequestMapping("/caseload", produces = [MediaType.APPLICATION_JSON_VALUE])
```
gurps-singh, line 29: **question:** ‏: A single caseload controller containing all the caseload functions or separate controllers for each caseload

andrewrlee: **suggestion:** Maybe we could just add them all to the existing `CaseloadController`?

PR #553, opened 2024-06-28
CVSL-1906 add recently approved endpoint

gurps-singh:
This PR is to add the last remaining endpoint for the approver caseload to get recently approved cases.


louisenorris (approved):
LGTM

PR #555, opened 2024-07-09
CVSL-1906 remove old endpoints and tests

gurps-singh:
This PR is the backend change to remove the two old endpoints used for the approver caseload.


andrewrlee (approved):
LGTM

PR #556, opened 2024-07-11
Cvsl 2003 cvl caseload move ca caseload from frontend to backend backend api

nishanthmahasamudram:
**note: CA(OMU) Caseload logic migration from frontend to backend**


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/CaCase.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  @Schema(description = "The prison identifier for the person on this licence", example = "A9999AA")
+  val prisonerNumber: String,
+
+  @Schema(description = "The details for the active supervising probation officer")
+  val probationPractitioner: ProbationPractitioner? = null,
+
+  @Schema(description = "The date on which the prisoner leaves custody", example = "30/11/2022")
+  @JsonFormat(pattern = "dd/MM/yyyy")
+  val releaseDate: LocalDate?,
```
nishanthmahasamudram, line 33: **thought: Passing back release date as LocalDate format instead of string and we can handle date not logic in backend.**

detekt-baseline.xml
```diff
@@ -2,6 +2,7 @@
 <SmellBaseline>
   <ManuallySuppressedIssues></ManuallySuppressedIssues>
   <CurrentIssues>
+    <ID>CyclomaticComplexMethod:LicenceStatus.kt$LicenceStatus.Companion$fun lookupLicenceEventByStatus(status: LicenceStatus): LicenceEventType</ID>
```
gurps-singh, line 5: **question:** ‏Whereas the other additions to this file are fine, this one should be addressed.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceSummary.kt
```diff
@@ -23,7 +23,7 @@ data class LicenceSummary(
   val licenceStatus: LicenceStatus,
 
   @Schema(description = "The prison nomis identifier for this offender", example = "A1234AA")
-  val nomisId: String?,
+  val nomisId: String,
```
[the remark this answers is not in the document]
gurps-singh: **question:** ‏Are we positive that there will no instances where the nomisId is null? Have a feeling this has come up before - tagging @SebNegahbanMoJ for historical context?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ManagedCase.kt
```diff
@@ -0,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.ManagedOffenderCrn
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.OffenderDetail
+
+@Schema(description = "Describes an CA(OMU) caseload")
```
gurps-singh, line 7: **suggestion:** ‏If these classes are only used in the creation of the caseload, better in include them as classes within the caseload service itself?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.isBreachOfTopUpSupervision
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.isEligibleEDS
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.isParoleEligible
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.isRecall
+import java.time.Clock
+import java.time.LocalDate
+
+@Service
+class CaCaseloadService(
+  val caseloadService: CaseloadService,
```
gurps-singh, line 44: **question:** ‏Any reason why this is not private?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+  private val clock: Clock,
+  private val probationSearchApiClient: ProbationSearchApiClient,
+  private val licenceService: LicenceService,
+  private val prisonApiClient: PrisonApiClient,
+  private val communityApiClient: CommunityApiClient,
+) {
+  fun getPrisonOmuCaseload(prisonCaseload: List<String>?, searchString: String?): CaCaseLoad {
```
gurps-singh, line 51: **question:** ‏These would always be set would they not if they are from the request from the frontend. Can't think of a situation where a request for caseload would not include these values - shouldn't be null?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ManagedCase.kt
nishanthmahasamudram: This is shared for COM caseload as well.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: Just followed **ApproverCaseloadService**

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+      )
+    }
+  }
+
+  fun findLatestLicenceSummary(licences: List<LicenceSummary>): LicenceSummary? {
+    if (licences.size == 1) {
+      return licences[0]
+    }
+    if (licences.any { it.licenceStatus == TIMED_OUT }) {
+      return licences.find { it.licenceStatus != TIMED_OUT }
```
gurps-singh, line 382: **question:** ‏This will return the first but do we need to check it is the most recent worked on licence if there are multiple licences or in order of preference of the licence status?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: Ya change this.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+      APPROVED,
+      NOT_STARTED,
+    ).contains(status) &&
+      noReleaseDates
+    val startDateInPast = licenceStartDate != null && status == APPROVED && licenceStartDate.isBefore(today)
+
+    return missingDates || startDateInPast
+  }
+
+  private fun selectReleaseDate(nomisRecord: Prisoner): LocalDate? {
```
gurps-singh, line 355: **suggestion:** ‏You can use the `PrisonerSearchFunctions` for this - specifically the `getLicenceStartDate` function.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+    if (eligibleExistingLicences.isEmpty() && eligibleNotStartedLicences.isEmpty()) {
+      return CaCaseLoad(cases = emptyList(), showAttentionNeededTab = false)
+    }
+
+    val cases = mapCasesToComs(eligibleExistingLicences + eligibleNotStartedLicences)
+
+    return buildCaseload(cases, searchString, "prison")
+  }
+
+  fun getProbationOmuCaseload(
```
gurps-singh, line 82: **suggestion:** ‏I would have possibly split this out into a separate service as this is quite a hefty service already.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: Ya, will add sort by while fetching licences this should solve the order.
**sortBy = "conditionalReleaseDate"**

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+
+  private fun getPrisonersApproachingRelease(
+    prisonCaseload: List<String>?,
+    overrideClock: Clock? = null,
+  ): Page<CaseloadItem>? {
+    val now = overrideClock ?: clock
+    val weeksToAdd: Long = 4
+    val today = LocalDate.now(now)
+    val todayPlusFourWeeks = LocalDate.now(now).plusWeeks(weeksToAdd)
+    return prisonCaseload?.toSet()?.let {
```
gurps-singh, line 367: **question:** ‏Why does this need to take the prison codes as a set?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceSummary.kt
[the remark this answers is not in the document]
SebNegahbanMoJ: There are no existing licences with a null NOMIS ID, and I believe it would likely cause significant issues long before this point if we managed to get one.
There also should be filtering to prevent us being able to create one in the first place, but I'm not sure it's there anymore. I inadvertently removed an explicit check during my frontend refactor, but it was still covered by [this filter](https://github.com/ministryofjustice/create-and-vary-a-licence/blob/main/server/services/lists/caCaseloadService.ts#L274), as `deliusRecord` would be `null` if there were no NOMIS ID.
That doesn't seem to be present anymore ([see here](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/556/files#diff-7a111cf0b4dbfa8329f4e11fbbe07c9da52636870204d209e9a59983ca3aac4aR425)) but it may be that Kotlin is doing some automatic null removal?
@nishanthmahasamudram Did the IDE suggest that the filter was unnecessary, hence why it was dropped?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
SebNegahbanMoJ: There should never be more than two licences if one of them is `TIMED_OUT`
One would be the NOT_STARTED/IN_PROGRESS licence that was timed out, and the optional other is the prison-created licence, if they've started it. This means that if there's a TIMED_OUT licence and more than one licence, we always want to return the other one

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+    } else if (isRecall(c)) {
+      // Offender is subject to an active recall - not clickable
+      licenceStatus = OOS_RECALL
+    } else if (c.cvlFields.isInHardStopPeriod) {
+      licenceStatus = TIMED_OUT
+    }
+
+    val com = c.deliusRecord?.managedOffenderCrn?.staff
+
+    val releaseDate = c.nomisRecord?.confirmedReleaseDate ?: c.nomisRecord?.conditionalReleaseDate
```
gurps-singh, line 202: **suggestion:** ‏Again, could use `getLicenceStartDate` against the `PrisonerSearchPrisoner` result here.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: Because of **prisonerSearchApiClient.searchPrisonersByReleaseDate** parma **prisonIds: Set<String>**.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+                staff = deliusRecord.offenderManagers.find { om -> om.active }?.staffDetail,
+              ),
+            ),
+          )
+        }
+        return@map ManagedCase(nomisRecord = prisoner, cvlFields = cvl)
+      }
+  }
+
+  private fun mapCasesToComs(casesToMap: List<CaCase>): List<CaCase> {
```
gurps-singh, line 428: **thought:** ‏This is quite confusing as to what is happening at first glance - here we are creating the caseload from the groupings with the relevant COMs where they exist or not - only by going through it bit by bit, you understand what is happening. Brain starting to hurt!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: will split this down.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+      .filter { offender -> offender.prisoner.conditionalReleaseDate != null }
+      .filter { offender ->
+        val status = offender.prisoner.status
+        return@filter (status != null && (status.startsWith("ACTIVE") || status == "INACTIVE TRN"))
+      }
+      .filter { offender ->
+        return@filter (
+          offender.prisoner.confirmedReleaseDate
+            ?: offender.prisoner.conditionalReleaseDate
+          )?.isAfter(LocalDate.now(clock)) == true
```
SebNegahbanMoJ, line 240: I don't think `isAfter` is appropriate here, unless Kotlin's implementation includes `today` in `isAfter` (I assume it doesn't)
We still need to return these cases if the release date is today

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
SebNegahbanMoJ: This bit is on me (the API is a recreating of the refactor that's happened in the UI)
In short, this is to reduce API calls and/or the size of said API calls to Delius. 
For cases where we have the COM staff code, we will also have their name, so don't need to call to Delius
For cases where we have the COM username, we should use that username to get the staff code and name from Delius
For cases where we have neither, we need to use the NOMIS ID to get the offender record, and get the COM details from that

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/MappaDetails.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation
+
+class MappaDetails(
```
SebNegahbanMoJ, line 3: There seem to be a lot of new classes for things that I don't believe we ever use. Do we need these?
I may be mistaken, but I believe you can leave out references to fields you don't need from API types, and it will still recognise the object as long as there's overlap with the existing fields.

In other words, you should be able to remove fields we don't use from `OffenderDetail.kt`, which would mean that we wouldn't need all of these new classes

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/CaseListUtils.kt
```diff
@@ -0,0 +1,66 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util
+
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.ManagedCase
```
gurps-singh, line 3: **suggestion:** ‏As mentioned above, this is part of the Eligibility Service which could be used instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/OffenderDetail.kt
```diff
@@ -4,4 +4,28 @@ data class OffenderDetail(
   val offenderId: Long,
   val otherIds: OtherIds,
   val offenderManagers: List<OffenderManager>,
+  val accessDenied: Boolean? = null,
```
gurps-singh, line 7: **note:** ‏I've been reluctant to all the fields in this response in the past going with an approach of only surfacing the fields required for the work. The problem with this is we now have 20+ extra fields on the response, the vast majority will not be used. I suppose that now we know exactly what we get from the endpoint. Maybe need to discuss a common approach to this?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/KeyValue.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation
+
+data class KeyValue(
```
gurps-singh, line 3: **thought:** ‏Could have used `Detail` for this?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadServiceTest.kt
```diff
@@ -0,0 +1,1306 @@
...
+    LicenceStatus.SUBMITTED,
+    LicenceStatus.IN_PROGRESS,
+    LicenceStatus.TIMED_OUT,
+    LicenceStatus.ACTIVE,
+  )
+
+  private val licenceQueryObject = LicenceQueryObject(
+    statusCodes = statuses,
+    prisonCodes = listOf("BAI"),
+//    sortBy = "conditionalReleaseDate",
```
gurps-singh, line 73: **nitpick:** ‏needs removing if not used.

gurps-singh:
As you can appreciate, this is quite a chunky PR. I've tried my best to go through and understand how everything lines up and made comments where I didn't quite get what is going on. I'm going to have another look tomorrow morning but would really appreciate a second pair of eyes on this (obvs not Seb or Nishanth seeing as you've worked on it) as it is quite a pivotal change and quite a few lines of code to get your head around.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: To make it clear I have added comments on it

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceSummary.kt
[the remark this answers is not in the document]
nishanthmahasamudram: Yes, It was IDE suggestion indeed.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: Now this was removed and using isEligibleForCvl from eligibility service

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,492 @@
...
+        ),
+      )
+    }
+
+    return caCaseList
+  }
+
+  private fun buildCaseload(cases: List<CaCase>, searchString: String?, view: String): CaCaseLoad {
+    val showAttentionNeededTab = cases.any { it.tabType == CaViewCasesTab.ATTENTION_NEEDED }
+    val searchResults = applySearch(searchString, cases)
```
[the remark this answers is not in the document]
nishanthmahasamudram: Ya, it's always better to move most of the possible functionality to backend.
So, it will be easy in the future if we want to make search functionality as common in the backend.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/MappaDetails.kt
nishanthmahasamudram: Yes, deleted unused models.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/OffenderDetail.kt
nishanthmahasamudram: removed all the unused fields.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/KeyValue.kt
nishanthmahasamudram: removed it as unused model.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadServiceTest.kt
nishanthmahasamudram: un commented and using it.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: I think it is kind of creating more logical complexity.
So better to be in one service.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -0,0 +1,462 @@
...
+
+  private fun createNotStartedLicenceForCase(cases: List<ManagedCase>): List<CaCase> = cases.map { c ->
+
+    // Default status (if not overridden below) will show the case as clickable on case lists
+    var licenceStatus = NOT_STARTED
+
+    if (isBreachOfTopUpSupervision(c)) {
+      // Imprisonment status indicates a breach of top up supervision order - not clickable (yet)
+      licenceStatus = OOS_BOTUS
+    } else if (isRecall(c)) {
```
nishanthmahasamudram, line 284: note: Not using the existing isRecallCase method in eligibility service as it is not returning false if recall flag is null.
Also it's a bit strange as we have to do all checks(**getIneligibilityReasons**) in order to execute **isRecallCase**.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/CaseloadItem.kt
```diff
@@ -37,7 +38,7 @@ data class CvlFields(
 @Schema(description = "An item in the caseload")
 data class CaseloadItem(
   @Schema(description = "Details about a prisoner")
-  val prisoner: Prisoner,
+  val prisoner: PrisonerSearchPrisoner,
```
gurps-singh, line 41: **question:** ‏Is this used anywhere else which may cause an unintended problem somewhere else (presumably not as everything is passing)?

nishanthmahasamudram: Ya it was used in one place and I have fixed it as well.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/GroupedByCom.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Describes an CA(OMU) case")
```
andrewrlee, line 5: This isn't a representation class - would just add as a private class to the service.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/LicenceSummary.kt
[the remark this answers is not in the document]
andrewrlee: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ManagedCase.kt
andrewrlee: **question:** is this exposed via the API? 
(It should be in the service layer only if not)

detekt.yml
```diff
@@ -107,7 +107,7 @@ complexity:
   CyclomaticComplexMethod:
     active: true
     threshold: 15
-    ignoreSingleWhenExpression: false
+    ignoreSingleWhenExpression: true
```
andrewrlee, line 110: **question**: why is this needed?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/GroupedByCom.kt
nishanthmahasamudram: We can add it to the service but not private, as it is used by public method.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
andrewrlee: this field on this class can be private though right? 

we shouldn't be accessing these kinds of fields directly but I don't think we are

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
andrewrlee: **Suggestion**: the representation should also have the field as `Set` so you don't need to convert in this level

detekt.yml
nishanthmahasamudram: Because of `CyclomaticComplexMethod` error caused by when statement with multiple clauses by detekt :
`<ID>CyclomaticComplexMethod:LicenceStatus.kt$LicenceStatus.Companion$fun lookupLicenceEventByStatus(status: LicenceStatus): LicenceEventType</ID>`

[https://github.com/detekt/detekt/issues/680]

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
nishanthmahasamudram: even if we do that it has to be converted down to list in **findLicencesMatchingCriteria** method.
But I did.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CaseloadService.kt
```diff
@@ -19,24 +20,32 @@ class CaseloadService(
...
+      .map { it.toCaseloadItem() }
 
   fun getPrisoner(nomisId: String) =
     getPrisonersByNumber(listOf(nomisId)).firstOrNull() ?: throw EntityNotFoundException(nomisId)
 
   private fun PrisonerSearchPrisoner.toCaseloadItem(): CaseloadItem {
     val sentenceDateHolder = this.toSentenceDateHolder()
     return CaseloadItem(
-      prisoner = this.toPrisoner(),
+      prisoner = this,
```
andrewrlee, line 38: **question**: why did these files change?

andrewrlee: oh.. you're trying to use it in the CaCaseloadService.. I would suggest just calling the client directly in your caseload service rather than trying to reuse this

nishanthmahasamudram: fixed it.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/CaseloadItem.kt
nishanthmahasamudram: fixed

gurps-singh (approved):

PR #557, opened 2024-07-15
CVSL-2025 remove prisoner search as can get the information from the licence summary

gurps-singh:
This PR is to refactor the logic in the Approver caseload to use licence data where available in preference over data from NOMIS and Delius. Upon doing so, turns out we don't need the call to the Prisoner search endpoint as it was only used to populate names and the NOMIS ID.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
```diff
@@ -45,22 +42,20 @@ class ApproverCaseloadService(
   private fun createApprovalCaseload(licences: List<LicenceSummaryApproverView>): List<ApprovalCase> {
     val nomisIds = licences.mapNotNull { it.nomisId }
     val deliusRecords = probationSearchApiClient.searchForPeopleByNomsNumber(nomisIds)
-    val deliusNomisIds = deliusRecords.mapNotNull { it.otherIds.nomsNumber }
-    val nomisRecords = prisonerSearchApiClient.searchPrisonersByNomisIds(deliusNomisIds)
 
-    val prisonerRecord: List<Triple<PrisonerSearchPrisoner, OffenderDetail?, LicenceSummaryApproverView?>> =
```
gurps-singh, line 51: **note:** ‏The current `ApprovalDetails` object was only using the NOMIS record for name and prisoner number.  I believe there are instances in Delius where certain people may not have a NOMIS ID but will have a CRN. Hence, got rid of the `PrisonerSearchPrisoner` record and used the Delius Records and the CRN to match licences.

SebNegahbanMoJ: **thought**: Agreed with the removal of the NOMIS call, but I'm not sure about the refactor of mapping by CRN. It _shouldn't_ be necessary for two reasons:

1. If memory serves, the approver caseload only shows cases that have had licences already created for them (ie doesn't show `NOT_STARTED`s). If this is the case, there should already be a link between NOMIS and Delius, otherwise I don't think the licence can be created. That said, it _is_ possible that a record has been merged, but in this case there would be a NOMIS ID but it would be incorrect.
2. [The call to fetch Delius records](https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/557/files#diff-855a1514b3f436b3a0588e2ddda17cef56d7cc70144917f6ed412fb2ee57785dR44) is done by NOMIS ID, meaning they must have a NOMIS ID to be returned. I would usually suggest also changing this call to search by CRN, but it may be that the `mapNotNull` is actually working somewhat as a filter to remove cases that _don't_ have a NOMIS ID on their Delius record, which is what you were trying to avoid in the first place!

I don't think it's doing any harm to change the map to be by CRN, as the previously-mentioned search by NOMIS ID will be removing any that are missing a NOMIS ID, and whatever we map them by from there should return the same result. My only concern would be edge cases. I have a vague recollection of a case where two NOMIS records were merged, but each ended up with a licence. I believe mapping by NOMIS ID would only return the correct (merged) record here, as the previous NOMIS ID would no longer match up to a Delius record. Mapping by CRN may return both though, as the CRN hadn't changed, so both licences would map to the same Delius record. This is incredibly rare though, and I could very well be wrong.

gurps-singh: Just checked the probation search endpoint - all is good. Will revert to using NOMIS ID instead of CRN.

SebNegahbanMoJ (approved):

PR #558, opened 2024-07-17
CVSL-1991: Add com case load endpoints and service.

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
justin-gardiner, line 1: This is mostly auto formatter updates, the only bit I added was 
fun Prisoner.toPrisonerSearchPrisoner()

justin-gardiner:
A couple of bits in here like ManagedCase I copied from Nish's branch to save both of us creating the same thing. I'll wait for him to merge his branch, so I handle any conflicts.

andrewrlee (approved):
LGTM

PR #559, opened 2024-07-19
CVSL-2029 add HDC licence kind

gurps-singh:
This PR is to introduce the HDC licence kind in the backend as part of a spike. This means, using the create licence endpoint, you can create a HDC licence.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceType.kt
```diff
@@ -14,5 +14,11 @@ enum class LicenceType {
       nomisRecord.topupSupervisionExpiryDate == null || nomisRecord.topupSupervisionExpiryDate <= nomisRecord.licenceExpiryDate -> AP
       else -> AP_PSS
     }
+
+    fun getHdcLicenceType(nomisRecord: PrisonerSearchPrisoner) = when {
```
gurps-singh, line 18: **note:** ‏Need to double check this with Shirley - I think based on our conversation this is right but also might be better to use the existing `getLicenceType` rule if that also makes sense.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -415,7 +415,7 @@ class LicenceService(
     try {
       val matchingLicences =
         licenceRepository.findAll(licenceQueryObject.toSpecification(), licenceQueryObject.getSort())
-      return matchingLicences.map { it.toSummary() }
+      return matchingLicences.filterNot { it.kind == LicenceKind.HDC }.map { it.toSummary() }
```
gurps-singh, line 418: **note:** ‏This is to currently exclude any HDC licence that may be created from surfacing to the caseload which use the `match` endpoint

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -141,6 +142,47 @@ class LicenceCreationService(
     return LicenceCreationResponse(createdLicence.id)
   }
 
+  @Transactional
+  fun createHdcLicence(prisonNumber: String): LicenceCreationResponse {
+    if (offenderHasLicenceInFlight(prisonNumber)) {
```
andrewrlee, line 147: this may change a bit in the future..

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceKind.kt
```diff
@@ -38,4 +38,13 @@ enum class LicenceKind(
     { LicenceDomainEventType.LICENCE_ACTIVATED },
     { LicenceDomainEventType.LICENCE_INACTIVATED },
   ),
+
+  HDC(
+    { IN_PROGRESS },
+    { LicenceEventType.HDC_CREATED },
+    { error("HDC licences can not be copied") },
```
andrewrlee, line 45: They probably will be able to but we don't have to worry about this now

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceType.kt
andrewrlee: **suggestion:** think it probably just makes sense to use `getLicenceType`

You can have PSS HDC licences it just happens very very rarely

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
andrewrlee: **suggestion**: think it's fine to not include this... it will hopefully reveal extra things we need to consider

andrewrlee (approved):
LGTM

PR #567, opened 2024-07-30
CVSL-1991: Refactor ComCaseloadService to return ComCase's instead of…

justin-gardiner:
… ManagedCase's.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ComCase.kt
```diff
@@ -0,0 +1,68 @@
...
+  @Schema(description = "The prison identifier for the person on this licence", example = "A9999AA")
+  val prisonerNumber: String?,
+
+  @Schema(description = "The date on which the prisoner leaves custody", example = "30/11/2022")
+  @JsonFormat(pattern = "dd/MM/yyyy")
+  val releaseDate: LocalDate?,
+
+  @Schema(description = "The date on which a list of returned COM cases has been sorted", example = "30/11/2022")
+  @JsonFormat(pattern = "dd/MM/yyyy")
+  val sortDate: LocalDate? = null,
```
gurps-singh, line 35: **question:** ‏For the Approver caseload, we removed this and sorted based on release date from the backend. Could this be done here or is `sortDate` used for something else?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
```diff
@@ -31,18 +34,18 @@ class ComCaseloadService(
     private const val PROBATION_SEARCH_BATCH_SIZE = 500
   }
 
-  fun getStaffCreateCaseload(deliusStaffIdentifier: Long): List<ManagedCase> {
```
gurps-singh, line 34: **question:** ‏Now that these are changed to the `ComCase` model, I presume these are used elsewhere otherwise we could remove the ManagedCase representation of the data?

justin-gardiner: unfortunately we still need ManagedCase as an intermediate type as we collect date from delius and nomis and combine before we can transform to ComCase later. It's also used by CaCaseloadService.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ComCase.kt
justin-gardiner: the sort is done on the backend, I've removed sortDate from here and added to the frontend as it is just used to populate an attribute on the table so say what it is sorted by.

gurps-singh (approved):

PR #571, opened 2024-08-06
CVSL-1911: Remove unnecessary code.

justin-gardiner:


nishanthmahasamudram (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
```diff
@@ -222,13 +221,12 @@ class ComCaseloadService(
           LicenceStatus.TIMED_OUT,
         )
       }
-      validLicenceStatus == true && !offender.nomisRecord.isBreachOfTopUpSupervision()
```
nishanthmahasamudram, line 225: Are we not checking **is breach of top up supervision case** anymore??

justin-gardiner: Yes it's already done by EligibilityService before we get here

nishanthmahasamudram: oh you are directly calling **getIneligibilityReasons**.
Ok.

PR #579, opened 2024-08-20
CVSL-2095 distinguish error when we detect requests that would create…

andrewrlee:
… duplicates

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
@@ -1,6 +1,5 @@
 package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
 
-import jakarta.validation.ValidationException
 import org.slf4j.LoggerFactory
 import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.stereotype.Service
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -208,10 +203,20 @@ class LicenceCreationService(
...
-    val inFlight =
+  private fun verifyNoInFlightLicence(nomsId: String) {
+    val inflightLicences =
       licenceRepository.findAllByNomsIdAndStatusCodeIn(nomsId, listOf(IN_PROGRESS, SUBMITTED, APPROVED, REJECTED))
-    return inFlight.isNotEmpty()
+
+    if (inflightLicences.isNotEmpty()) {
+      val currentInflightLicence = when {
+        inflightLicences.size == 1 -> inflightLicences.first()
+        else -> inflightLicences.first { it.statusCode != APPROVED }
```
andrewrlee, line 213: if there's more than one inflight licence then there will only be two licences and one would be approved. 
We wouldn't want to choose the APPROVED one

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationServiceTest.kt
```diff
@@ -459,22 +463,54 @@ class LicenceCreationServiceTest {
...
+
+      val exception = assertThrows<ResourceAlreadyExistsException> {
+        service.createLicence(prisonNumber)
+      }
+
+      assertThat(exception)
+        .isInstanceOf(ResourceAlreadyExistsException::class.java)
+        .withFailMessage("A licence already exists for this person (IN_PROGRESS, SUBMITTED, APPROVED or REJECTED)")
+
+      assertThat(exception.existingResourceId).isEqualTo(unapprovedLicence.id)
```
andrewrlee, line 512: we ignore the approved licence and choose the unapproved one

nishanthmahasamudram (approved):
LGTM

PR #584, opened 2024-08-29
CVSL-2033 add endpoint to call hdc api for curfew data

gurps-singh:
This PR is to add an endpoint to retrieve HDC licence data for a licence. The idea is that this endpoint will be used in the frontend where we currently render the PDF for a licence where its kind is HDC. This will mean that there is only an additional endpoint called when necessary and not for every single licence.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/HdcController.kt
```diff
@@ -0,0 +1,63 @@
...
+      ),
+      ApiResponse(
+        responseCode = "400",
+        description = "Bad request, request body must be valid",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "401",
+        description = "Unauthorised, requires a valid Oauth2 token",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
```
andrewrlee, line 50: do you need a 404 response?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@Service
+class HdcService(
+  private val hdcApiClient: HdcApiClient,
+) {
+
+  @Transactional
+  fun getHdcLicenceData(bookingId: String?): HdcLicenceData? {
+    if (bookingId == null) {
+      return null
```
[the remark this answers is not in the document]
andrewrlee: actually this can never be null, it's a path param isn't it?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    val licenceData = this.hdcApiClient.getByBookingId(bookingId.toLong())
```
andrewrlee, line 19: **suggestion:** This should probably be bound to a long in the controller parameter.
This would then be converted to a 400 bad request if its not parseable

(Generally worth pushing any conversion/marshalling to the boundaries of the application - fail fast rather than potentially having to deal with bad data in midst of business logic)

andrewrlee: Worth thinking of the error handling [here](https://github.com/ministryofjustice/hmpps-hdc-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppshdcapi/licences/LicenceService.kt#L18). 

Ideally the hdc-api would return a 404 if there is no licence and that would be propagated through this service

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/hdc/HdcApiClient.kt
```diff
@@ -4,19 +4,18 @@ import org.springframework.beans.factory.annotation.Qualifier
...
 
-  fun getByBookingId(bookingId: Long): HdcLicence? {
+  fun getByBookingId(bookingId: Long): HdcLicenceData? {
     return hdcApiWebClient
       .get()
       .uri("/licence/hdc/$bookingId")
       .accept(MediaType.APPLICATION_JSON)
       .retrieve()
-      .bodyToMono(typeReference<HdcLicence>())
+      .bodyToMono(typeReference<HdcLicenceData>())
```
andrewrlee, line 18: **suggestion:** worth thinking about the error handling [here](https://github.com/ministryofjustice/hmpps-hdc-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppshdcapi/licences/LicenceService.kt#L18).

Ideally the hdc api would 404 and this service would propagate the 404 response through to the ui.

andrewrlee: if the backend api returns a 404 this would throw a 500 response.. might be worth having a look at `coerce404ResponseToNull(it)` and create an equivalent `propagate404Response(it)`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/hdc/HdcApiClient.kt
```diff
@@ -1,21 +1,41 @@
...
       .uri("/licence/hdc/$bookingId")
       .accept(MediaType.APPLICATION_JSON)
       .retrieve()
-      .bodyToMono(typeReference<HdcLicenceData>())
+      .bodyToMono(HdcLicenceData::class.java)
+      .onErrorResume { propagate404Response(it, bookingId) }
       .block()
   }
+
+  private fun <API_RESPONSE_BODY_TYPE> propagate404Response(exception: Throwable, bookingId: Long): Mono<API_RESPONSE_BODY_TYPE> =
```
gurps-singh, line 31: **note:** ‏I'm hoping this is what you meant by propagating - if there is an exception and it's a NOT_FOUND, this will throw a 404 and not 500 - integration test included.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/HdcController.kt
```diff
@@ -0,0 +1,68 @@
...
+  @Operation(
+    summary = "Get curfew details from HDC.",
+    description = "Get curfew details from HDC. Requires ROLE_CVL_ADMIN.",
+    security = [SecurityRequirement(name = "ROLE_CVL_ADMIN")],
+  )
+  @ApiResponses(
+    value = [
+      ApiResponse(
+        responseCode = "200",
+        description = "The HDC curfew information was retrieved",
```
andrewrlee, line 40: **suggestion(non-blocking):** probably need to add content with the response type in there

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -0,0 +1,23 @@
...
+class HdcService(
+  private val hdcApiClient: HdcApiClient,
+) {
+
+  @Transactional
+  fun getHdcLicenceData(bookingId: Long): HdcLicenceData? {
+    val licenceData = this.hdcApiClient.getByBookingId(bookingId)
+
+    return HdcLicenceData(
+      curfewAddress = licenceData?.curfewAddress,
```
andrewrlee, line 18: **suggestion:** I guess if you're handling the 404 then licenceData could be non nullable?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/HdcControllerTest.kt
```diff
@@ -0,0 +1,115 @@
...
+@ExtendWith(SpringExtension::class)
+@ActiveProfiles("test")
+@WebMvcTest(controllers = [HdcController::class])
+@AutoConfigureMockMvc(addFilters = false)
+@ContextConfiguration(classes = [HdcController::class])
+@WebAppConfiguration
+class HdcControllerTest {
+
+  @MockBean
+  private lateinit var hdcService: HdcService
```
andrewrlee, line 42: **suggestion:** rather than mock out the service it might be better to let it call wiremock so it's a proper end to end test.

Also then you'd be able to test out the 404 behaviour.

andrewrlee: ah.. see you've got a separate integration test for it - that's fine then!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/hdc/HdcLicenceData.kt
```diff
@@ -1,6 +1,6 @@
 package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.hdc
 
-data class HdcLicence(
+data class HdcLicenceData(
```
andrewrlee, line 3: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/HdcController.kt
gurps-singh: Done

andrewrlee (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
gurps-singh: Done

andrewrlee (approved):
LGTM

PR #588, opened 2024-09-12
CVSL-2033 add missing HDC kind to prison approver service

gurps-singh:
Pr to address a bug we found during testing where the "Recently Approved" tab was not working as expected as forgot to add the HDC licence kind to logic when finding the original licence for variation.


louisenorris (approved):
LGTM

PR #590, opened 2024-09-20
PI-2450: Replace calls to community API with calls to cvl and delius

pmcphee77:
This PR is an effort to move away from the deprecated Community API and replace calls to your dedicated cvl-and-delius API. Need to work out how we test this in the test environments before we release to prod.
**detekt** seems to be complaining due to the number of constructor params (which i havent changed, just renamed one)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -193,15 +194,15 @@ class CaCaseloadService(
...
 
     val caCaseListWithStaffUsername = cases.withStaffUsername.map { caCase ->
       val com = coms[caCase.probationPractitioner?.staffUsername?.lowercase()]
       if (com != null) {
         caCase.copy(
           probationPractitioner = ProbationPractitioner(
-            staffCode = com.staffCode,
-            name = com.staff?.let { "${it.forenames} ${it.surname}" },
+            staffCode = com.code,
+            name = com.name?.let { "${it.forename} ${it.surname}" },
```
andrewrlee, line 205: is there a reason this doesn't use `fullName`?

pmcphee77: Great spot! Ive changed this. Also fixed detekt

andrewrlee (approved):
LGTM

andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #595, opened 2024-10-08
CVSL-2151 Condition 4a refactor

SebNegahbanMoJ:
Condition 4a has essentially had both a text change, and a split between the existing condition and a new condition.
We don't currently have functionality to handle a text change and split, so we instead decided it would be better to model it as a condition deletion and replacement by two new conditions.
This was approved by the policy team.


andrewrlee (approved):
LGTM

PR #598, opened 2024-10-28
CVSL-2238 Adding missed tweaks to policy condition wording

SebNegahbanMoJ:
QA review spotted some changes to condition content that was missed.


andrewrlee (approved):
LGTM

PR #601, opened 2024-11-01
CVSL-2245 Changing getLicenceType to set AP as default if dates are missing

SebNegahbanMoJ:
There are a whole bunch of support requests that stem from sentence dates being missing when a licence is created.
The most impactful issue of these is having to change the licence template, as this is a wholly manual process that requires live data fixing.
As `AP` is the most common licence type and `PSS` is relatively niche, it makes the most sense to have `AP` be the default when all dates are missing.


louisenorris (approved):
LGTM

PR #602, opened 2024-11-05
alee/CVSL-2266 fix errors when staff with teams search for offenders

andrewrlee:
We could probably still do with some message to the user that they have no teams


justin-gardiner (approved):
LGTM

PR #608, opened 2024-11-20
CVSL-2270 add integration tests for hdc pss error

gurps-singh:
This PR is to test whether an error is thrown when a HDC PSS licence is created.


andrewrlee (approved):
LGTM

PR #616, opened 2024-12-18
CVSL-2329 replace LocalTime schema with string

gurps-singh:
This PR is to fix a LocalTime issue where it was being unmarshalled as a string but as an object:

```
{
    "hour": 0,
    "minute": 0,
    "nano": 0,
    "second": 0
}
```
 The fix includes creating a new `Schema<LocalTime>` object, setting its type to "string", its format to "HH:mm", and providing an example value. Then I have replaced the default schema for `LocalTime` with this custom schema.
 
 Before:
 
Screenshot 2024-12-18 at 09 57 18

 After:

Screenshot 2024-12-18 at 09 55 09


andrewrlee (approved):
LGTM

PR #621, opened 2024-12-27
CVSL-2329 change first night curfew times to LocalTime instead of string

gurps-singh:
This PR is to update the type for First Night curfew times in the API. The HDC API is now using LocalTime so this matches the type.


andrewrlee (approved):
LGTM

PR #624, opened 2025-01-03
CVSL-2399 enabling scheduled downtime

andrewrlee:
This will shut down pods between 10pm - 6:30am UTC on weekdays and all day on weekends. 
6:30am was chosen as the RDS startup happens between 6am and 6:30am.


nishanthmahasamudram (approved):
LGTM

PR #625, opened 2025-01-06
CVSL-2335 Using new LSD calculation for licence creation

SebNegahbanMoJ:
Includes necessary changes to eligibility service


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -86,7 +85,7 @@ class EligibilityService(
   }
 
   private fun hasReleaseDateInTheFuture(): EligibilityCheck = early@{
-    val releaseDate = it.getLicenceStartDate() ?: return@early true
+    val releaseDate = it.conditionalReleaseDate ?: return@early true
```
SebNegahbanMoJ, line 88: Calculating licence start date here caused at least one extra query to NOMIS per case (n queries). As such, it was decided to just check CRD, as we get this for free and should represent eligibility well enough (in combination with the other rules)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PrisonerSearchService.kt
```diff
@@ -39,15 +40,30 @@ class PrisonerSearchService(
...
+      )
+    }
+
+    val prisonersWithoutLicences = cvlSearchRecords.filter { searchRecord -> searchRecord.licence == null }.map {
+      it.prisonerSearchPrisoner
+    }
+    val licenceStartDates = releaseDateService.getLicenceStartDates(prisonersWithoutLicences)
+
+    val searchResults = cvlSearchRecords.mapNotNull {
+      val licenceStartDate = licenceStartDates[it.prisonerSearchPrisoner?.prisonerNumber]
```
gurps-singh, line 63: The nullable `prisonerSearchPrisoner` here is throwing me a bit - so we would have a Delius record and no NOMIS record? I always thought the reverse was possible (NOMIS record but no Delius record) but not this way round?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ReleaseDateService.kt
```diff
@@ -107,6 +107,22 @@ class ReleaseDateService(
     }
   }
 
+  fun getLicenceStartDates(records: List<PrisonerSearchPrisoner?>): Map<String, LocalDate?> {
+    val prisoners = records.filterNotNull()
+    val iS91BookingIds: List<Long> = iS91DeterminationService.getIS91AndExtraditionBookingIds(prisoners)
+    return prisoners.associate {
```
gurps-singh, line 113: Nice!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PrisonerSearchService.kt
SebNegahbanMoJ: There _shouldn't_ ever be a time that there's no NOMIS record, but in reality there can be a disconnect between Delius and NOMIS, whether that's an incomplete link between CRN and noms ID, or a duplicate NOMIS record has been removed and references not been updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/CvlProbationSearchRecord.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.Licence
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonerSearchPrisoner
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.CaseloadResult
+
+@Schema(description = "CVL Probation Search Record")
+data class CvlProbationSearchRecord(
+  @Schema(description = "Delius caseload record")
+  val caseloadResult: CaseloadResult,
```
andrewrlee, line 11: mmh.. whats this API representation for? 

(If it's just an intermediate representation then it can be a private inner class in the service)

SebNegahbanMoJ: Ah, you've hit the nail on the head there
I thought this was the blanket pattern of creating a representation in Kotlin, I thought it was a bit odd providing schema annotations etc!

I'll refactor it into a private inner class

andrewrlee (approved):
LGTM

PR #631, opened 2025-01-21
CVSL-2269 Updating licence repository queries to use LSD

SebNegahbanMoJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
```diff
@@ -192,7 +192,7 @@ interface LicenceRepository : JpaRepository<Licence, Long>, JpaSpecificationExec
       SELECT l
         FROM Licence l
         WHERE (l.statusCode IN (uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.APPROVED, uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.SUBMITTED, uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.IN_PROGRESS)
-                AND l.conditionalReleaseDate IS NULL AND l.actualReleaseDate IS NULL)
+                AND l.licenceStartDate IS NULL)
```
SebNegahbanMoJ, line 195: This shouldn't change the logic for licences with the old LSD calculation, as that route used `LSD = ARD || CRD`, so if CRD and ARD were both null, LSD would be null too. It's also the only way that LSD could have been null, so it should be behaviourally identical.

andrewrlee (approved):
LGTM

PR #636, opened 2025-01-30
CVSL-2269 Adding new LSD logic to COM caseload

SebNegahbanMoJ:
Ensures that the release date is correctly calculated for NOT_STARTED cases.

Also changed logic to filter out cases without NOMIS records as they wouldn't be eligible for CVL anyway, and even if one did sneak through somehow, it would blow up on `LicenceType.getLicenceType(case.nomisRecord!!)`
This could _potentially_ impact post-release cases if something happened to the PoP's NOMIS record, but that would be an error case that arguably we shouldn't account for anyway.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
@@ -12,11 +12,13 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.DeliusRecor
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.LicenceService
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.ManagedCase
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.PrisonerSearchService
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.ReleaseDateService
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.conditions.convertToTitleCase
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.DeliusApiClient
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.ManagedOffenderCrn
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.ProbationSearchApiClient
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.fullName
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.toPrisonerSearchPrisoner
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceKind
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
```diff
@@ -125,10 +128,13 @@ class ComCaseloadService(
   fun mapOffendersToLicences(cases: List<ManagedCase>): List<ManagedCase> {
     val nomisIdList = cases.mapNotNull { offender -> offender.nomisRecord?.prisonerNumber }
     val existingLicences: List<LicenceSummary> = findExistingLicences(nomisIdList)
+    val casesToLicences = cases.filter { it.nomisRecord != null }.associateWith {
+      existingLicences.filter { licence -> it.nomisRecord!!.prisonerNumber == licence.nomisId }
```
andrewrlee, line 132: could tidy this up slightly by:
```kt
val existingLicences = findExistingLicences(nomisIdList).groupingBy { it.nomisId }
val casesToLicences = cases
       .filter { it.nomisRecord != null }
       .associateWith { existingLicences[ it.nomisRecord!!.prisonerNumber ] }

File: src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt [hunk 2]
```diff
@@ -210,7 +216,7 @@ class ComCaseloadService(
     managedOffenders.filter { offender ->
       offender.nomisRecord?.status?.startsWith("ACTIVE") == true || offender.nomisRecord?.status == "INACTIVE TRN"
     }.filter { offender ->
-      val releaseDate = offender.nomisRecord?.confirmedReleaseDate ?: offender.nomisRecord?.conditionalReleaseDate
+      val releaseDate = offender.licences.find { licence -> offender.licences.size == 1 || licence.licenceStatus != LicenceStatus.ACTIVE }?.releaseDate
```
andrewrlee, line 219: not a fan of this weird short circuiting of find.. this is saying return the ACTIVE one if there's only a single licence is that right?  which would then be filtered out by the next step?

andrewrlee: (Out of scope of this work - I think identifying the single licence upfront generally leads to loads of simplification when building these caseload. I see we've got this short circuiting in a few places)

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
SebNegahbanMoJ: Honestly I'm not a huge fan of it either and I considered trying to find a neater and/or more explicit `find`, but as you said this is used elsewhere and is known to work, and I thought increasing the scope of this ticket further was a bad idea! 😅 

The logic of it is `If there's one licence, that must be the one we want. If it's not, take the non-ACTIVE one.`
The reason this works is that if there are multiple pre-release licences, we can pick any of them to check the release date as they will all respond to NOMIS updates and will have the same dates, so it doesn't matter which the `find` returns. If there are multiple post-release licences, the non-ACTIVE one will be the variation, which will be most up-to-date (though it also shouldn't matter here, but sometimes support overrides are only done to variations by mistake).

andrewrlee (approved):
LGTM

PR #643, opened 2025-02-06
CVSL-2478 consistently batch calls at the api layer

andrewrlee:
We apply batching inconsistently and at different levels- this moves batching down and applies it consistently.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
@@ -33,10 +33,6 @@ class ComCaseloadService(
   private val probationSearchApiClient: ProbationSearchApiClient,
   private val releaseDateService: ReleaseDateService,
 ) {
-  companion object {
-    private const val PROBATION_SEARCH_BATCH_SIZE = 500
-    private const val PRISONER_SEARCH_BATCH_SIZE = 500
-  }
 
   fun getStaffCreateCaseload(deliusStaffIdentifier: Long): List<ComCase> {
     val managedOffenders = deliusApiClient.getManagedOffenders(deliusStaffIdentifier)
@@ -84,10 +80,7 @@ class ComCaseloadService(
 
   private fun mapManagedOffenderRecordToOffenderDetail(caseload: List<ManagedOffenderCrn>): List<DeliusRecord> {
     val crns = caseload.map { case -> case.crn }
-    val batchedCrns = crns.chunked(PROBATION_SEARCH_BATCH_SIZE)
-    val batchedOffenders = batchedCrns.map { batch -> probationSearchApiClient.getOffendersByCrn(batch) }
-
-    val offenders = batchedOffenders.flatten()
+    val offenders = probationSearchApiClient.getOffendersByCrn(crns)
     return offenders.map { offender ->
       DeliusRecord(offender, caseload.find { case -> case.crn == offender.otherIds.crn }!!)
     }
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/Batching.kt
```diff
@@ -0,0 +1,18 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util
+
+object Batching {
+  fun interface Gatherer<ID, RESULT> {
+    fun gather(ids: List<ID>): List<RESULT>?
+  }
```
andrewrlee, line 6: created an interface so I could mock it

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/Batching.kt
```diff
@@ -0,0 +1,18 @@
...
+    size: Int,
+    ids: Collection<ID>,
+    gatherer: Gatherer<ID, RESULT>,
+  ): List<RESULT> {
+    if (ids.isEmpty()) return emptyList()
+    val batches = ids.chunked(size)
+    val results = batches.map { batch -> gatherer.gather(batch) ?: emptyList() }
+    return results.flatten()
+  }
+}
```
andrewrlee, line 18: generic batching function

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadServiceTest.kt
```diff
@@ -1116,52 +1126,6 @@ class ComCaseloadServiceTest {
...
-      deliusApiClient.getManagedOffendersByTeam(selectedTeam),
-    ).thenReturn(managedOffenders)
-
-    service.getTeamVaryCaseload(listOf("team A", "team B"), listOf(selectedTeam))
-    verify(deliusApiClient).getManagedOffendersByTeam(selectedTeam)
-    verify(probationSearchApiClient, times(3)).getOffendersByCrn(any())
-  }
-
-  @Test
-  fun `it should batch calls to the get prisoners by prisoner number service`() {
```
andrewrlee, line 1140: tests not needed as done in the client now

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadService.kt
```diff
@@ -96,10 +89,8 @@ class ComCaseloadService(
   private fun pairDeliusRecordsWithNomis(managedOffenders: List<DeliusRecord>): List<ManagedCase> {
     val caseloadNomisIds = managedOffenders.filter { offender -> offender.offenderDetail.otherIds.nomsNumber != null }
       .mapNotNull { offender -> offender.offenderDetail.otherIds.nomsNumber }
-    val batchedNomisIds = caseloadNomisIds.chunked(PRISONER_SEARCH_BATCH_SIZE)
-    val batchedNomisRecords = batchedNomisIds.map { batch -> caseloadService.getPrisonersByNumber(batch) }
+    val nomisRecords = caseloadService.getPrisonersByNumber(caseloadNomisIds)
```
andrewrlee, line 92: this method used to pass batches down to two separate calls - one to get prisoners and the other to get court outcomes, both the underlying calls do batching so don't need to worry about batching at this level anymore

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/DeliusApiClient.kt
```diff
@@ -51,36 +53,29 @@ class DeliusApiClient(@Qualifier("oauthDeliusApiClient") val communityApiClient:
...
+  fun getStaffEmails(crns: List<String>) =
+    batchRequests(STAFF_EMAIL_BATCH, crns) { batch ->
       communityApiClient
         .post()
         .uri("/probation-case/responsible-community-manager")
         .bodyValue(batch)
         .accept(MediaType.APPLICATION_JSON)
         .retrieve()
         .bodyToMono(typeReference<List<StaffEmail>>())
-        .block() ?: error("Unexpected null response from API")
```
louisenorris, line 66: How come we don't want to through an error here anymore? I guess it would just throw an error anyway if an error occurred?

louisenorris (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/probation/DeliusApiClient.kt
andrewrlee: It was just to make it consistent with the others - the batching function handles nulls by coercing them into an empty list.

I think it makes sense to handle these consistently as there's no reason why the behaviour should differ based on the endpoint. I went with coerce to null rather than error, as that was the most common pattern - there is an argument to consistently error if we encounter nulls though.

gurps-singh: Around this point, there are a couple of places where we propagate 404's to 404 rather 500 (eg in the `HdcApiClient`) - would be good to get consistent on that as well throughout our codebases. In terms of nulls, how would we ever know then if we coerce whether a genuine error with null data has occurred?

gurps-singh (approved):

PR #645, opened 2025-02-10
CVSL-2478 Filter to ensure release dates are in range

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/promptingCom/PromptComListBuilder.kt
```diff
@@ -97,25 +94,25 @@ class PromptComListBuilder(
 
   fun excludeInHardStop(cases: List<CaseWithEmailAndStartDate>) = cases.filter { it.isNotInHardStop() }
 
+  fun excludeOutOfRangeDates(cases: List<CaseWithEmailAndStartDate>, startDate: LocalDate, endDate: LocalDate) =
+    cases.filter { (_, _, releaseDate) -> releaseDate in startDate..endDate }
+
```
andrewrlee, line 99: this is the only change^ - the rest is formatting

PR #646, opened 2025-02-10
CVSL-2269 adding new LSD calculation to CVLFields for caseload items

SebNegahbanMoJ:


andrewrlee (approved):
LGTM - sweet so no extra calculation

PR #649, opened 2025-02-13
CVSL-2485 add test coverage

gurps-singh:
This PR is to add a test to confirm that HDC licences do not time out following a sentence date update which moves them into the hard stop window. The job and sentence date changes already work to not time out the HDC licences but this is just a test to cover the scenario.


louisenorris (approved):
LGTM

PR #652, opened 2025-02-13
CVSL-2521 add is-91 status endpoint

louisenorris:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/OffenderController.kt
```diff
@@ -202,4 +202,53 @@ class OffenderController(
...
+  )
+  @ApiResponses(
+    value = [
+      ApiResponse(
+        responseCode = "200",
+        description = "a boolean for IS-91 status",
+        content = [
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = String::class),
```
gurps-singh, line 225: If it is a boolean, should it not return a Boolean class type?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/PrisonerSearchService.kt
```diff
@@ -94,6 +95,14 @@ class PrisonerSearchService(
     return reasons + hdcReasonIfPresent
   }
 
+  fun getIS91Status(prisonNumber: String): Boolean {
+    val prisoners = prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(prisonNumber))
+    if (prisoners.size != 1) {
+      error("Found ${prisoners.size} prisoners for: $prisonNumber")
```
gurps-singh, line 101: Not too sure we should just error as this would crash the service - unless we have a better way to handle this? Can see we do the same in the getIneligibilityReasons endpoint so it's fine.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/OffenderController.kt
louisenorris: Thank you! Corrected 😄

gurps-singh (approved):

PR #653, opened 2025-02-14
CVSL-2505 exclude HDC licences from the current deactivate job

gurps-singh:
This PR adds logic to exclude HDC licences from the current deactivation job. HDC licences will be deactivated using their own custom job.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/DeactivateLicencesService.kt
```diff
@@ -34,7 +35,7 @@ class DeactivateLicencesService(
   @Transactional
   fun deactivateLicences() {
     log.info("Job deactivateLicencesJob started")
-    val licencesToDeactivate = licenceRepository.getDraftLicencesPassedReleaseDate()
+    val licencesToDeactivate = licenceRepository.getDraftLicencesPassedReleaseDate().filterNot { it.kind == HDC }
```
gurps-singh, line 38: **question:** ‏Unsure whether it is better pushing this down to the query level rather than here so do `kind != 'HDC'`? Thoughts?

SebNegahbanMoJ: I think I'd probably say it's better to be on the query level, if only to reduce the number of loops/ functions being called on the return data.
I would assume it's ever-so-slightly more performant to put a constraint on the query than it is to filter the returned data

andrewrlee: yes, makes sense to tweak the query. 
(checking it's not used elsewhere)

andrewrlee (approved):
LGTM

PR #639, opened 2025-02-04
CVSL-2269 Adding new LSD calculation logic to SentenceDateChange event handler

SebNegahbanMoJ:

PR #654, opened 2025-02-17
CVSL-2269 Fixing prison api sentence detail response

SebNegahbanMoJ:
PR includes the revert of #651, which itself was a revert of #639, so reimplements the sentence date change handler alterations

Also includes the fix to make #639 work


andrewrlee (approved):
LGTM

PR #668, opened 2025-02-25
CVSL-2559 Moving over to HDC service for support and com prompting

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -61,5 +66,7 @@ class HdcService(
       val approvedForHdc = approvedIds.contains(bookingId)
       return (kind == HDC && approvedForHdc) || (kind != HDC && !approvedForHdc)
     }
+
+    fun isApprovedForHdc(bookingId: Long) = approvedIds.contains(bookingId)
```
gurps-singh, line 70: Slight threw me that this function and the new addition are both called `isApprovedForHdc` but can see that one is a function for the service and one for the HdcStatuses

gurps-singh (approved):

PR #671, opened 2025-02-26
CVSL-2559 Moving over CA caseload service to use HDC service

andrewrlee:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadServiceTest.kt
```diff
@@ -739,80 +742,7 @@ class CaCaseloadServiceTest {
...
-              ),
-              isDueForEarlyRelease = false,
-              lastWorkedOnBy = null,
-            ),
-          ),
-        )
-      }
-
-      @Test
-      fun `should not filter out cases with an approved HDC licence but no HDCED`() {
```
andrewrlee, line 784: test not needed as this is covered by HDC service behaviour

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -67,8 +64,22 @@ class HdcService(
...
+     */
     fun canBeSeenByCom(kind: LicenceKind?, bookingId: Long) = isValidByKind(kind, bookingId)
 
+    /**
+     * For CA:
+     * Always show started cases regardless of HDC status.
+     * TODO: will need to fix when we start to show prospective HDC cases in caselist
+     * If licence hasn't been started, only show !approved licences.
+     */
+    fun canUnstartedCaseBeSeenByCa(bookingId: Long) = !isApprovedForHdc(bookingId)
```
andrewrlee, line 81: This is the current logic, feels like we should consistently apply eligibility logic across started and unstarted cases for COM, CA and other caseloads too.. right?

PR #673, opened 2025-02-27
CVSL-2559 Fixes for HDC cases

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadSearchService.kt
```diff
@@ -139,7 +141,8 @@ class ComCaseloadSearchService(
     if (prisoners.isEmpty()) {
       return this
     }
-    val prisonersForHdcCheck = this.filter { it.licenceStatus == NOT_STARTED }.mapNotNull { prisoners[it.nomisId] }
+    val prisonersForHdcCheck =
+      this.filter { PRE_RELEASE_STATUSES.contains(it.licenceStatus) }.mapNotNull { prisoners[it.nomisId] }
```
andrewrlee, line 145: I think this could be `it.isOnProbation == false` 

i.e, the opposite of below:
`this.filter { it.isOnProbation == true || hdcStatuses.canBeSeenByCom(it.kind, it.bookingId!!) }`

I think the only difference is TIMED_OUT licences, but if you have a TIMED_OUT licence that is either a "not approved" HDC cases or an HDC approved CRD case then I think that should be filtered out - it's a product of the case rather than the licence status

(The tests pass either way currently as I'm not checking TIMED_OUT licences)

andrewrlee: it'd be pretty easy to add this behaviour now actually... just need to filter out in the HdcService and bookings coming up to CRD.

This would make filter out all HDC licences and reveal CRD licences... would need thorough testing though.

SebNegahbanMoJ: > I think the only difference is TIMED_OUT licences: we would filter out CRD licences in the hardstop period if they're approved for HDC.
I think I could go either way on this, I can't think of a scenario where it would matter hugely as hard stop doesn't apply to HDC

> I guess technically we should prevent HDC approval status from filtering out CRD licences at CRD-9, yes they're approved but they'll never be released on HDC. We obviously have never done this previously
Agreed 

> would need thorough testing though.
Absolutely!

PR #685, opened 2025-03-10
CVSL-2592 add missing logic to search to handle missing LSD

gurps-singh:
This PR applies consistent logic across COM caseload and COM search. We now check for the missing release date in search as well as checking that any in flight (excluding active) licences have a release date in the future


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/ComIntegrationTest.kt
```diff
@@ -66,7 +66,7 @@ class ComIntegrationTest : IntegrationTestBase() {
           "Staff Surname",
           "A01B02C",
           "Test Team",
```
gurps-singh, line 68: Now we are doing date comparisons when checking for past release dates, need to adjust this seed data.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadSearchServiceTest.kt
```diff
@@ -635,6 +639,7 @@ class ComCaseloadSearchServiceTest {
     whenever(prisonerSearchApiClient.searchPrisonersByNomisIds(any())).thenReturn(listOf(aPrisonerSearchResult))
     whenever(eligibilityService.isEligibleForCvl(any())).thenReturn(true)
     whenever(hdcService.getHdcStatus(any())).thenReturn(HdcStatuses(emptySet()))
```
gurps-singh, line 641: This was missing a call to get licence start dates as it tests for not started licence creation.

PR #686, opened 2025-03-12
CVSL-2647 Removing past PED check from new licence start date calculations

SebNegahbanMoJ:
The past PED check has resulted in several support issues reporting that the LSD has been calculated incorrectly.

From discussing with Laura, we believe that any cases intended to be caught by this check would also have an APD, which would make them ineligible for CVL. As such, this check is redundant and is causing issues for users, so is being removed.


andrewrlee (approved):
LGTM

PR #687, opened 2025-03-13
CVSL-2593 Preventing edit and submit for ineligible licences

SebNegahbanMoJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -395,10 +398,25 @@ class LicenceService(
       ?: throw ValidationException("Staff with username $username not found")
 
     val updatedLicence = when (licenceEntity) {
-      is CrdLicence -> licenceEntity.submit(submitter as CommunityOffenderManager)
+      is CrdLicence -> {
+        if (!licenceHolderIsEligible(licenceEntity.nomsId)) {
```
andrewrlee, line 402: mmh.. could just move this check up front and do it once?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -395,10 +398,25 @@ class LicenceService(
...
+          throw ValidationException("Can only submit cases eligible for CVL")
```
andrewrlee, line 403: 400 response is fine.. could do a 409 conflict but it's probably not worth the faff considering this is the internal API and we don't need to distinguish between the two types of failure.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -1099,6 +1121,15 @@ class LicenceService(
     isDueToBeReleasedInTheNextTwoWorkingDays = releaseDateService.isDueToBeReleasedInTheNextTwoWorkingDays(this),
   )
 
+  private fun licenceHolderIsEligible(nomisId: String?): Boolean {
+    if (nomisId == null) {
+      return false
+    }
+
+    val prisoner = prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(nomisId)).first()
+    return eligibilityService.isEligibleForCvl(prisoner)
```
andrewrlee, line 1130: could potentially log a warn if submitting an ineligible?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
SebNegahbanMoJ: I did initially, but as we're not blocking it for `VariationLicence`s, I thought it best to avoid unnecessary API calls to prisoner search

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
SebNegahbanMoJ: Absolutely can do, but as the result of the check failing is to throw an error in all scenarios, I thought we would be able to check the logs for the errors instead. Appreciate this means that if this function is ever used for another purpose, it wouldn't necessarily log

andrewrlee: well maybe we could change this to `assertCaseEligibile` and throw here?.. which would address that concern and would mean we'd get rid of all the nesting/cruft from above.  

(I don't think we use the term licence holder anywhere else so not sure it makes sense to start using it here?)

SebNegahbanMoJ: I agree it's odd, but I wanted to differentiate it from `isEligibleForCvl` which assesses eligibility for cases regardless of if they have a licence already or not. This function is for people who already have licences, and for whom we also don't do HDC checks.
Happy to rename it though and throw the error from here. The only issue is that we'd lose which action (eg edit or submit) caused the error to be thrown, though we should be able to figure that out from the support query and the error trace

andrewrlee (approved):
LGTM

PR #689, opened 2025-03-19
CVSL-2596 change model to include county and adjust names of variables

gurps-singh:
This PR is to adjust the model which returns a curfew address from the HDC API. This is to match what will be from the HDC API for curfew addresses.


louisenorris (approved):
LGTM

PR #691, opened 2025-03-26
CVSL-2596 add hdc curfew address entity

gurps-singh:
This PR allows us to store curfew addresses in CVL so we can re-render the associated licences without relying on state in HDC.


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/HdcIntegrationTest.kt
```diff
@@ -121,6 +122,104 @@ class HdcIntegrationTest : IntegrationTestBase() {
       )
     }
```
gurps-singh, line 124: **note:** ‏was missing some test coverage on getting saved curfew data so added this test in for both curfew hours and curfew address.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcServiceTest.kt
```diff
@@ -315,7 +328,7 @@ class HdcServiceTest {
       listOf(
         EntityHdcCurfewTimes(
           1L,
```
gurps-singh, line 330: **note:** ‏changed this to refer to `aLicenceEntity` as this already calls `createHdcLicence()`

louisenorris (approved):
LGTM

PR #693, opened 2025-03-31
CVSL-2648 save curfew address against hdc licence on creation

louisenorris:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -187,6 +189,11 @@ class LicenceCreationService(
     val standardConditions = licencePolicyService.getStandardConditionsForLicence(createdLicence)
     standardConditionRepository.saveAllAndFlush(standardConditions)
 
+    val curfewAddress = hdcService.getHdcLicenceData(createdLicence.id)?.curfewAddress
```
andrewrlee, line 192: * Rename `isEligibleForHdcLicence`  to `checkEligibleForHdcLicence`
* Call the api to get the curfew address up front (line 158-159)
* Pass in the curfew address then we can remove the api call from the check method
* refer to curfewAddress variable down here. 
* `curfewAddress!!` down here rather than a null check

louisenorris: Not sure we can move the api call to get the curfew address to before the  newly named `checkEligibleForHdcLicence` as we don't have access to the `createdLicence.id` parameter for this until we save and flush the licence 🤔

louisenorris: Adding a new function to get address by booking Id to the HDC service to resolve above issue

andrewrlee (approved):
LGTM

PR #696, opened 2025-04-03
CVSL-2671 add null curfew times check to checkEligibleForHdcLicence

louisenorris:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -43,6 +44,12 @@ class HdcService(
     return licenceData.curfewAddress
   }
 
+  @Transactional
+  fun getCurfewTimesByBookingId(bookingId: Long): List<HdcCurfewTimes>? {
+    val licenceData = this.hdcApiClient.getByBookingId(bookingId)
```
andrewrlee, line 49: best to not do two calls to get the same information.. could just return the hdc info as a whole and then pass that through to the eligibility service

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -38,9 +37,9 @@ class HdcService(
   fun isApprovedForHdc(bookingId: Long, hdced: LocalDate?) = if (hdced == null) false else prisonApiClient.getHdcStatus(bookingId).isApproved()
 
   @Transactional
-  fun getCurfewAddressByBookingId(bookingId: Long): HdcCurfewAddress? {
+  fun getHdcLicenceDataByBookingId(bookingId: Long): HdcLicenceData? {
```
louisenorris, line 40: Renamed this as now retrieving `HdcLicenceData` in one client call rather than for each curfew times and curfew address. We have a function below named `getHdcLicenceData` which retrieved `HdcLicenceData` with a `licenceId` parameter. Should this be renamed for consistency to `getHdcLicenceDataByLicenceId` or is it ok to leave as is?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
louisenorris: Now only doing one API call for this as suggested - thank you!

andrewrlee (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -86,9 +85,13 @@ class HdcService(
       error("HDC licence for ${nomisRecord.prisonerNumber} could not be created as they are not approved for HDC")
     }
 
-    if (curfewAddress == null) {
+    if (hdcLicenceData?.curfewAddress == null) {
       error("HDC licence for ${nomisRecord.prisonerNumber} could not be created as there is no curfew address")
     }
+
+    if (hdcLicenceData.curfewTimes == null) {
```
andrewrlee, line 92: I guess this will change when we capture curfew times for HDC licences within CVL

PR #698, opened 2025-04-08
CVSL-2651 Adding HDC Variation creation endpoint

SebNegahbanMoJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/Licence.kt
```diff
@@ -134,6 +134,8 @@ abstract class Licence(
   @ManyToOne(fetch = FetchType.EAGER)
   @JoinColumn(name = "updated_by_id", nullable = true)
   var updatedBy: Staff? = null,
+
+  var variationOfId: Long? = null,
```
andrewrlee, line 138: Shouldn't this be on the subclass too?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/HdcLicence.kt
```diff
@@ -244,7 +244,7 @@ data class HdcLicence(
   override val updatedByUsername: String? = null,
 
   @Schema(description = "The curfew times for this licence")
-  val curfewTimes: HdcCurfewTimes? = null,
```
andrewrlee, line 247: mmh.. how is this not broken on the frontend?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceFactory.kt
```diff
@@ -253,6 +254,60 @@ object LicenceFactory {
...
   )
 
+  fun createHdcVariation(licence: HdcLicence, creator: CommunityOffenderManager): HdcVariationLicence {
+    with(licence) {
+      return HdcVariationLicence(
+        id = -1,
+        typeCode = this.typeCode,
+        version = this.version,
+        statusCode = VARIATION_IN_PROGRESS,
+        variationOfId = licence.id,
```
andrewrlee, line 264: meh.. to be consistent this could be `this.id` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -104,10 +106,7 @@ fun transformToLicenceSummary(
   isDueForEarlyRelease = isDueForEarlyRelease,
   isDueToBeReleasedInTheNextTwoWorkingDays = isDueToBeReleasedInTheNextTwoWorkingDays,
   updatedByFullName = licence.getUpdatedByFullName(),
-  homeDetentionCurfewActualDate = when (licence) {
-    is HdcLicence -> licence.homeDetentionCurfewActualDate
-    else -> null
-  },
+  homeDetentionCurfewActualDate = if (licence.isHdcLicence()) licence.homeDetentionCurfewActualDate else null,
```
andrewrlee, line 109: guess this check isn't technically required as the SentenceDateHolder would return null for none-variations. But think this is fine

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/HdcLicence.kt
andrewrlee: and will this change break the frontend?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -150,6 +153,13 @@ class LicenceService(
       conditionSubmissionStatus = conditionSubmissionStatus,
     )
 
+    is HdcVariationLicence -> toHdcVariation(
+      licence = licence,
+      earliestReleaseDate = earliestReleaseDate,
+      isEligibleForEarlyRelease = isEligibleForEarlyRelease,
+      conditionSubmissionStatus = conditionSubmissionStatus,
+    )
```
andrewrlee, line 161: This file seems to contain a lot of changes that are covered by other tickets (and not related to the creation endpoint) which there's no test coverage for?

SebNegahbanMoJ: This was something I mentioned in standup, though it may have been a day that you were on leave.
I was slightly confused about the scope of this ticket vs the "endpoints" ticket. Whether this was purely create-only, meaning the UI journey would be entirely inaccessible (as `HdcVariationLicence` would be completely unhandled except for in the `createVariation` endpoint), or whether this should be the functional user-journey of creating an HDC variation. Eg create -> answer SPO/VLO questions -> Submit

I went with the latter, with the view that this work would be next to untestable via the UI and would make the endpoint work significantly larger without these additional bits. If you'd rather them be in separate PRs, I can break them down though.

SebNegahbanMoJ: As for testing, you're right, I completely missed them somehow. I'll get some in.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/HdcLicence.kt
SebNegahbanMoJ: > will this change break the frontend?

I don't believe so, no. The front end expects an array of curfew times to be returned, I believe this is an existing bug that is being fixed. That said, I'm not sure how it wasn't broken before. Looking into that now.

Below is a screenshot of the HDC Curfew details in my local with my changes, it all seems to be rendering as expected to me
image

andrewrlee: magic javascript

PR #703, opened 2025-04-15
CVSL-2767 Updating PP endpoints for HDC variations

SebNegahbanMoJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Licence.kt
```diff
@@ -21,7 +21,7 @@ object LicenceKinds {
 
 @Schema(
   description = "Describes a licence within this service, A discriminator exists to distinguish between different types of licence",
-  oneOf = [CrdLicence::class, VariationLicence::class, HardStopLicence::class, HdcLicence::class],
+  oneOf = [CrdLicence::class, VariationLicence::class, HardStopLicence::class, HdcLicence::class, HdcVariationLicence::class],
```
andrewrlee, line 24: good spot!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/LicencePolicyController.kt
```diff
@@ -142,20 +142,20 @@ class LicencePolicyController(
...
-    return previousLicence?.let { licencePolicyService.policyByVersion(it) }
+  fun getPolicyChanges(currentLicence: ModelVariation, parentLicenceId: Long?, version: String): List<LicenceConditionChanges> {
+    if (parentLicenceId === null) return emptyList()
+    val parentLicence = licenceService.getLicenceById(parentLicenceId)
+
+    return licencePolicyService.compareLicenceWithPolicy(
+      currentLicence,
+      parentLicence,
+      version,
+    )
```
andrewrlee, line 159: (this is all a bit messy which is why its difficult to work with - ideally we'd sort it out properly and move out into a service, fix the potential circular dependency and  work on the licence entity rather than the model object but all for another time!)

andrewrlee (approved):
LGTM

PR #707, opened 2025-04-23
PI-2987 Migrate Delius search endpoints to integration service

marcus-bcl:
This change removes the dependency on Probation Offender Search, and instead uses the existing endpoints on the Delius integration API to find a person's probation details by CRN or NOMIS id.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ApproverCaseloadService.kt
```diff
@@ -43,99 +38,59 @@ class ApproverCaseloadService(
...
+        releaseDate = licenceSummary?.licenceStartDate,
+        urgentApproval = licenceSummary?.isDueToBeReleasedInTheNextTwoWorkingDays,
+        approvedBy = licenceSummary?.approvedByName,
+        approvedOn = licenceSummary?.approvedDate,
+        isDueForEarlyRelease = licenceSummary?.isDueForEarlyRelease,
+        kind = licenceSummary?.kind,
       )
     }.sortedWith(compareBy(nullsFirst()) { it.releaseDate })
-
+    println(approvalCases)
```
andrewrlee, line 66: println

marcus-bcl: Oops good spot - thanks, removed!

PR #711, opened 2025-05-07
CVSL-2595 Use consistent set of sentence dates

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceFunctions.kt
```diff
@@ -20,22 +20,45 @@ data class SentenceChanges(
...
+  sentenceStartDate = sentenceStartDate,
+  sentenceEndDate = sentenceExpiryOverrideDate ?: sentenceExpiryDate,
+  licenceExpiryDate = licenceExpiryOverrideDate ?: licenceExpiryDate,
+  topupSupervisionStartDate = topupSupervisionStartDate,
+  topupSupervisionExpiryDate = topupSupervisionExpiryOverrideDate ?: topupSupervisionExpiryDate,
+  postRecallReleaseDate = postRecallReleaseOverrideDate ?: postRecallReleaseDate,
+  homeDetentionCurfewEligibilityDate = homeDetentionCurfewEligibilityDate,
+  homeDetentionCurfewActualDate = homeDetentionCurfewActualDate,
+  homeDetentionCurfewEndDate = homeDetentionCurfewEndDate,
+)
```
andrewrlee, line 35: This logic is lifted from [here](https://github.com/ministryofjustice/create-and-vary-a-licence/blob/a43bc81223260f8068517087c4bca37f08edbd3b/server/listeners/eventHandlers/prisonEvents/datesChangedEventHandler.ts#L81-L105)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceFunctions.kt
```diff
@@ -20,22 +20,45 @@ data class SentenceChanges(
...
+  val ledChanged = licenceExpiryDate.hasChanged(sentenceDates.licenceExpiryDate)
+  val sedChanged = sentenceEndDate.hasChanged(sentenceDates.sentenceEndDate)
+  val tussdChanged = topupSupervisionStartDate.hasChanged(sentenceDates.topupSupervisionStartDate)
+  val tusedChanged = topupSupervisionExpiryDate.hasChanged(sentenceDates.topupSupervisionExpiryDate)
+  val prrdChanged = postRecallReleaseDate.hasChanged(sentenceDates.postRecallReleaseDate)
+  val hdcadChanged =
+    this is HdcLicence && homeDetentionCurfewActualDate.hasChanged(sentenceDates.homeDetentionCurfewActualDate)
   val hdcEndDateChanged =
-    this is HdcLicence && nullableDatesDiffer(newSentence.homeDetentionCurfewEndDate, this.homeDetentionCurfewEndDate)
+    this is HdcLicence && homeDetentionCurfewEndDate.hasChanged(sentenceDates.homeDetentionCurfewEndDate)
```
SebNegahbanMoJ, line 61: I know this is existing logic and it's not currently something we need to worry about as in context it's only ever used for pre-release licences, but it may be worth us remembering that this won't be true for HDC Variations, so we can't rely on this function to tell us if the HDC dates have changed in NOMIS for a post-release case. There's no scenario where we need to right now that I can think of, but I thought it worth raising so it wasn't just in my head!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceFunctions.kt
SebNegahbanMoJ: This is delightfully neater

SebNegahbanMoJ (approved):

PR #713, opened 2025-05-12
CVSL-2837: Add some logging to COM allocation.

justin-gardiner:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/StaffIntegrationTest.kt
```diff
@@ -49,6 +55,51 @@ class StaffIntegrationTest : IntegrationTestBase() {
     }
   }
 
+  // This will fail randomly due more than one thread finding no existing COM and so
```
justin-gardiner, line 58: Disabled for now and may remove depending on where we do handling of processing more that one update for COM allocation concurrently.

andrewrlee (approved):
LGTM

PR #717, opened 2025-05-13
CVSL-2815 add update for hdc curfew times

louisenorris:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcServiceTest.kt
```diff
@@ -365,9 +388,57 @@ class HdcServiceTest {
     }
   }
 
+  @Nested
+  inner class `update curfew times` {
```
andrewrlee, line 392: we should really be adding integration tests for new api endpoints

Ewan-Donovan (approved):

PR #723, opened 2025-05-16
CVSL-2873 remove hdc specific event types

louisenorris:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceKind.kt
```diff
@@ -53,8 +53,8 @@ enum class LicenceKind(
   HDC_VARIATION(
     { VARIATION_IN_PROGRESS },
     { error("HDC variation licences are only copied from existing HDC licences") },
-    { LicenceEventType.HDC_VARIATION_CREATED },
-    { LicenceEventType.HDC_VARIATION_SUBMITTED },
+    { LicenceEventType.VARIATION_CREATED },
+    { LicenceEventType.VARIATION_SUBMITTED },
     { LicenceDomainEventType.HDC_LICENCE_VARIATION_ACTIVATED },
     { LicenceDomainEventType.HDC_LICENCE_VARIATION_INACTIVATED },
```
louisenorris, line 59: Think `LicenceDomainEventType`s remain with HDC prefix but highlighting this to check that's correct.

PR #733, opened 2025-05-21
CVSL-2890 Sanitising test data

SebNegahbanMoJ:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/LicenceControllerTest.kt
```diff
@@ -634,7 +634,7 @@ class LicenceControllerTest {
       licenceStatus = LicenceStatus.IN_PROGRESS,
       nomisId = "A1234AA",
       forename = "Bob",
-      surname = "Mortimer",
+      surname = "Bloggs",
```
louisenorris, line 637: Don't want to ask an annoying question 😅  but are we def ok to use `Bloggs` and common names like `John` `Joe` `Jim` below? Is it safer to use `Some Person` or other def non names?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/support/LicenceTypeOverrideServiceTest.kt
```diff
@@ -315,7 +315,7 @@ class LicenceTypeOverrideServiceTest {
       argumentCaptor<AuditEvent>().apply {
         verify(auditEventRepository, times(1)).saveAndFlush(capture())
         assertThat(firstValue.detail).isEqualTo("ID 1 type PSS status IN_PROGRESS version 1.1")
-        assertThat(firstValue.username).isEqualTo("smills")
+        assertThat(firstValue.username).isEqualTo("tcom")
         assertThat(firstValue.summary).isEqualTo("Licence type overridden for John Smith: from AP_PSS to PSS: Test change to PSS")
```
louisenorris, line 319: Do we care about `John Smith`? 😅

src/test/resources/test_data/seed-audit-events.sql
```diff
@@ -17,12 +17,12 @@ detail,
...
+ (8,3,'2022-09-10 03:00:00.000','CVL_OMU','Omu User','USER_EVENT','Licence viewed for john doe','ID 3 type AP status APPROVED version 2.0', null),
  (9,5,'2022-09-10 03:00:00.000','CVL_OMU','Omu User','USER_EVENT','Licence viewed for prisoner five','ID 5 type AP status SUBMITTED version 2.0', null),
  (10,5,'2022-09-10 03:00:00.000','CVL_OMU','Omu User','USER_EVENT','Licence viewed for prisoner five','ID 5 type AP status APPROVED version 2.0', null),
  (11,6,'2022-09-10 03:00:00.000','CVL_OMU','Omu User','USER_EVENT','Licence viewed for prisoner six','ID 6 type AP status SUBMITTED version 2.0', null),
  (12,6,'2022-09-10 03:00:00.000','CVL_OMU','Omu User','USER_EVENT','Licence viewed for prisoner six','ID 6 type AP status APPROVED version 2.0', null),
- (13,4,'2022-09-10 03:00:00.000','CVL_OMU','Omu User','USER_EVENT','Licence viewed for bob pint','ID 4 type AP status ACTIVE version 2.0', null);
+ (13,4,'2022-09-10 03:00:00.000','CVL_OMU','Omu User','USER_EVENT','Licence viewed for bob bobson','ID 4 type AP status ACTIVE version 2.0', null);
```
louisenorris, line 28: Maybe we want to continue with using `prisoner [instert number]` for consistency in this file and remove common names?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/LicenceControllerTest.kt
SebNegahbanMoJ: I think our general take is that if a name has become an idiom for "an unknown person" (eg John Smith, Joe Bloggs, and John/Jane Doe) that it's probably fine, as it's clearly not a reference to an actual person.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/support/LicenceTypeOverrideServiceTest.kt
SebNegahbanMoJ: Same as above, John Smith being an idiom, I think it's fine to leave

src/test/resources/test_data/seed-audit-events.sql
SebNegahbanMoJ: I can do, I was trying to change as little as possible but it also wouldn't be a huge issue to do so

SebNegahbanMoJ: Done! I've also replaced "Prisoner" with "Person", as I believe "Prisoner" is generally to be avoided

andrewrlee (approved):
LGTM

PR #758, opened 2025-06-06
CVSL-2936-Distinguishing-conditions-that-require-electronic-monitorin…

nishanthmahasamudram:
…g-responses


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/policy/AdditionalConditionAp.kt
```diff
@@ -11,4 +11,5 @@ data class AdditionalConditionAp(
   val subtext: String? = null,
   override val type: String? = null,
   override val skippable: Boolean? = false,
+  val requiresEmResponse: Boolean = false,
```
andrewrlee, line 14: can we call this `requiresElectronicMonitoringResponse` ? 

(just might be a bit confusing for future maintainers)

PR #767, opened 2025-06-13
CVSL-2973-Adding-feature-flag-electronicMonitoringResponseHandling-to…

nishanthmahasamudram:
…-helm-config-and-enum-to-the-licence-model-representations'


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/ElectronicMonitoringProvider.kt
```diff
@@ -11,7 +10,4 @@ data class ElectronicMonitoringProvider(
 
   @Schema(description = "Programme Name of the licence", example = "Off Some Road")
   val programmeName: String? = null,
```
nishanthmahasamudram, line 12: moved to licence object.

gurps-singh (approved):

PR #770, opened 2025-06-18
Cvsl 2869 prisoner update event

justin-gardiner:


gurps-singh (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/wiremock/DeliusMockServer.kt
```diff
@@ -82,30 +82,10 @@ class DeliusMockServer : WireMockServer(8093) {
     )
   }
 
-  fun stubGetOffenderManagers() {
+  fun stubGetOffenderManagerError(crn: String = "X12345") {
```
justin-gardiner, line 85: I should have deleted this as not used, will do in a follow up

PR #772, opened 2025-06-18
CVSL-2913 add prison code and description for new location column in …

gurps-singh:
…ca search

This PR is to add prison code and description to the CA Case model as this is needed for the new location column for CA search in the frontend.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/CaCase.kt
```diff
@@ -59,4 +59,10 @@ data class CaCase(
 
   @Schema(description = "Is the licence in the hard stop period? (Within two working days of release)")
   val isInHardStopPeriod: Boolean = false,
+
+  @Schema(description = "The agency code where this offender resides or was released from", example = "MDI")
+  val prisonCode: String? = null,
```
andrewrlee, line 64: should this be nullable? required = true?

andrewrlee (approved):
LGTM

PR #775, opened 2025-06-20
CVSL-2913 change from string to list where a user may work at multipl…

gurps-singh:
…e prisons

This PR is to change the request body for a CA search as a user may be attached/work at multiple prisons and as such have multiple prison codes within which they want to search for a case.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/request/PrisonUserSearchRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/request/PrisonUserSearchRequest.kt
@@ -1,7 +1,7 @@
 package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.model.request
 
 import io.swagger.v3.oas.annotations.media.Schema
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/request/PrisonUserSearchRequest.kt
```diff
@@ -1,15 +1,15 @@
...
 
   @Schema(description = "The query the user wishes to search for (e.g. CRN, name, NOMIS ID)", example = "Joe Bloggs")
   val query: String,
 
-  @Schema(description = "The prison caseloads of the prison staff member", example = "BAI")
-  @field:NotNull
-  val prisonCaseload: String,
+  @Schema(description = "The prison caseloads of the prison staff member", example = "[BAI]")
+  @field:NotEmpty
+  val prisonCaseload: List<String>,
```
andrewrlee, line 14: Guess this should be`prisonCaseLoads`, and you could make this a `Set` so you don't need to transform lower down?

PR #776, opened 2025-06-24
CVSL-3010 Refactoring activation of IS91 licences to remove reliance on CRD

SebNegahbanMoJ:
The "new" Licence Start Date calculation handles IS91 cases and assigns the relevant LSD. This means the activation logic should be using the new LSD, which simplifies the logic and removes reliance on CRD.


AledWEvans (approved):
LGTM

PR #779, opened 2025-06-25
CVSL-2992 add Caija's DPD and change config

gurps-singh:
This PR is to add Caija's DPD to the backend. This is necessary JSON containing details of the query we want to run and the report format we need.


louisenorris (approved):
LGTM

PR #801, opened 2025-07-04
CVSL-3017 Recalls | Exclude Hard stop logic for PRRD Cases

AledWEvans:
Have a look guys.

I separated the getHard stop logic in a sense for cases and licences, because the logic for CRD not having a hardStopDate if in the past. But when there viewing a licence in the future and CRD is in the past and it has been approved what do we do then??? Put this PR up so we can think in more detail about this, you may have already and know the answer!


AledWEvans:
This might not be happing

PR #804, opened 2025-07-07
CVSL-3027 Recalls | change email template to pass through CRD

AledWEvans:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/NotifyService.kt
```diff
@@ -63,14 +63,14 @@ class NotifyService(
...
+    date: LocalDate?,
   ) {
-    if (emailAddress != null && crd != null) {
+    if (emailAddress != null && date != null) {
       val values: Map<String, String> = mapOf(
         "prisonerFirstName" to firstName,
         "prisonerLastName" to lastName,
         "prisonerNumber" to (prisonerNumber ?: "unknown"),
-        "crd" to crd.format(dateFormat),
+        "crd" to date.format(dateFormat),
```
AledWEvans, line 73: Was not sure if I should change this to key to LSD as the template would break in prod!

andrewrlee: we can remove the old key after the template change has been applied in prod. 
You can send extra data and it won't complain - it will just error if there is not enough

AledWEvans: good call

andrewrlee (approved):
LGTM

PR #810, opened 2025-07-08
CVSL-3031: When determining licence start dates batch calls to determ…

justin-gardiner:
…ine IS91 cases. Fix potential NPE in determineLicenceKind.


andrewrlee (approved):
LGTM

PR #822, opened 2025-07-11
Cvsl 3022 error handling

ryantk:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/documents/DocumentApiClient.kt
```diff
@@ -30,9 +30,10 @@ class DocumentApiClient(@Qualifier("oauthDocumentApiClient") val documentApiClie
     .accept(MediaType.APPLICATION_JSON)
     .retrieve()
     .onStatus(HttpStatusCode::isError) { response ->
-      val errorResponse = response.bodyToMono<Map<String, String>>()
-      error("Error during uploading document (status=${response.statusCode()}, body=$errorResponse)")
+      response.bodyToMono<Map<String, String>>().map { json ->
```
andrewrlee, line 33: if you're not using the structure of the payload, wondering if it'd make sense to make this less typesafe.. I'm wondering if it would fail to log properly on an nginx 502 error page for instance as it wouldn't be able to parse the html into a map?

andrewrlee: and thinking jackson might fail to parse the json unless its strictly conforms to a `Map<String, String>`?

(could be wrong!)

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/documents/DocumentApiClientTest.kt
```diff
@@ -49,34 +49,47 @@ class DocumentApiClientTest {
...
-      file = file.contentAsByteArray,
-      documentUuid = UUID.fromString("e2487a03-7cf9-4a9c-85e4-1d51efd7b3f1"),
-      documentType = DocumentType.EXCLUSION_ZONE_MAP,
-      metadata = mapOf("aKey" to "1", "bKey" to "2", "cKey" to "3"),
-    )
+    val result = uploadDocument()
 
-    assertEquals(result?.documentUuid, UUID.fromString("e2487a03-7cf9-4a9c-85e4-1d51efd7b3f1"))
-    assertEquals(result?.documentType, DocumentType.EXCLUSION_ZONE_MAP)
+    assertEquals(result.documentUuid, UUID.fromString("e2487a03-7cf9-4a9c-85e4-1d51efd7b3f1"))
```
andrewrlee, line 56: we tend to use assertJ assertions rather than the native junit.. leads to slightly nicer error messages and it's easier to just statically import a single thing which is then contextual based on the passed in type. 
`assertThat(1).isGreaterThan(2)` as it's number
`assertThat("bob").contain("ob")` as it's a string.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/documents/DocumentApiClientTest.kt
```diff
@@ -49,34 +49,47 @@ class DocumentApiClientTest {
...
+      ),
+      exception.message,
+    )
+  }
+
+  @Test
+  fun `Throws an exception when the document is not successful and no response could be parsed`() {
+    givenDocumentApiRespondsWith(status = 401, responseBody = "")
+
+    val exception = assertThrows<IllegalStateException> { uploadDocument() }
```
andrewrlee, line 81: equivalent for assertJ is [here](https://github.com/assertj/assertj-examples/blob/72c339f0f72fd4a6e903c25af2d3f093702f3e4c/assertions-examples/src/test/java/org/assertj/examples/ExceptionAssertionsExamples.java#L268)

andrewrlee (approved):
LGTM

gurps-singh (approved):

PR #824, opened 2025-07-11
CVSL-3015: Update isDueToEarlyRelease for recall licences which are e…

justin-gardiner:
…xcluded from early release schemes.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -403,25 +400,6 @@ class CaCaseloadService(
     }
   }
 
-  private fun PrisonerSearchPrisoner.toCaseloadItem(licenceStartDate: LocalDate?): CaseloadItem {
```
justin-gardiner, line 406: There were two identical methods for this so removed this duplicate.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/dates/ReleaseDateService.kt
```diff
@@ -130,6 +137,28 @@ class ReleaseDateService(
     }
   }
 
+  fun determineLicenceKind(sentenceDateHolder: SentenceDateHolder): LicenceKind {
```
justin-gardiner, line 140: moved here as it's needed for isDueForEarlyRelease and it's also dependent on a method in the class to get hard stop date

justin-gardiner:
Closing PR following. Sebs comment. Will leave branch here for now.

PR #826, opened 2025-07-14
CVSL-3083 follow up work from the labels ticket moved the logic into …

AledWEvans:
Follow up work for the release labels, moved the various methods into a util class for maintainability and visibility.

The old code was all over the place and duplicating functionality.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
@@ -18,6 +18,7 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.Pris
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonerSearchPrisoner
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.CaseloadResult
 import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.probation.fullName
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.util.ReleaseDateLabelFactory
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/util/ReleaseDateLabelFactory.kt
```diff
@@ -0,0 +1,60 @@
...
+  )
+
+  @JvmStatic
+  fun fromPrisonerSearch(licenceStartDate: LocalDate?, offender: PrisonerSearchPrisoner): String = getLabel(
+    releaseDate = licenceStartDate,
+    confirmedReleaseDate = offender.confirmedReleaseDate,
+    postRecallDate = offender.postRecallReleaseDate,
+    hdcReleaseDate = offender.homeDetentionCurfewActualDate,
+  )
+}
```
AledWEvans, line 60: Might be odd that these are static, as I would rather use a spring component but we are forced to do this because it is called from a extension function.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -757,41 +758,32 @@ fun CaseloadResult.transformToModelFoundProbationRecord(
   isInHardStopPeriod: Boolean,
   isDueForEarlyRelease: Boolean,
   isDueToBeReleasedInTheNextTwoWorkingDays: Boolean,
-): ModelFoundProbationRecord {
-  val hdcad = if (licence.isHdcLicence()) licence.homeDetentionCurfewActualDate else null
```
louisenorris, line 761: Guess the `hdcad` is not needed anymore? 😅 doesn't look like it was returned but just flagging.

louisenorris (approved):
LGTM

PR #827, opened 2025-07-15
CVSL-3022: Api client and service layer for document downloads

ryantk:
Added API client and service method for downloading document files.
Not currently plumbed in to anything.
Removed some excessive repetition of the uuid in one of the tests.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/documents/DocumentApiClient.kt
```diff
@@ -12,6 +12,21 @@ import java.util.UUID
 
 @Service
 class DocumentApiClient(@Qualifier("oauthDocumentApiClient") val documentApiClient: WebClient) {
+
+  // Swagger documentation: https://document-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/document-controller/downloadDocumentFile
+  fun downloadDocumentFile(documentUuid: UUID): ByteArray = documentApiClient.get()
```
AledWEvans, line 17: /*
 This is a multi-line comment.
 It spans multiple lines.
*/
May be add comments like this

PR #828, opened 2025-07-16
CVSL-3070 Allowing PRRD licences to enter the hard stop period

SebNegahbanMoJ:
Also includes a couple of minor refactors


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadServiceTest.kt
```diff
@@ -167,6 +174,12 @@ class CaCaseloadServiceTest {
     inner class `in the hard stop period` {
       @Test
       fun `Sets NOT_STARTED licences to TIMED_OUT when in the hard stop period`() {
+        whenever(caseloadService.prisonerToCaseloadItem(any(), anyOrNull())).thenReturn(
+          TestData.caseLoadItem().copy(
```
AledWEvans, line 178: prisonerToCaseloadItem(any(), anyOrNull())) this is the bit I change not sure if it should be null best check

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -243,6 +246,26 @@ class LicenceCreationService(
...
+    }
+
+    val hardstopDate = releaseDateService.getHardStopDate(nomisRecord.toSentenceDateHolder(null, kind))
+    if (hardstopDate != null && hardstopDate <= today) {
+      kind = LicenceKind.HARD_STOP
+    }
+
+    return kind
+  }
+
```
AledWEvans, line 268: this code looks familiar... did we do this in the case load service. could we call this from the case load service so we don't want to duplicate,... it dose not belong in the case load service it belongs here... I will go and check....
`caseloadService.determineLicenceKind(it)`
here it is, yes remove from case load service and use your code

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CaseloadService.kt
```diff
@@ -60,19 +61,4 @@ class CaseloadService(
       ),
     )
   }
-
-  fun determineLicenceKind(nomisRecord: PrisonerSearchPrisoner): LicenceKind {
```
AledWEvans, line 64: great you have removed...

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
AledWEvans: I can see you have removed I will shut up

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/SentenceDateHolder.kt
```diff
@@ -1,11 +1,16 @@
...
 interface SentenceDateHolder {
+  val kind: LicenceKind
   val licenceStartDate: LocalDate?
   val conditionalReleaseDate: LocalDate?
   val actualReleaseDate: LocalDate?
   val homeDetentionCurfewActualDate: LocalDate?
     get() = null
+  val postRecallReleaseDate: LocalDate?
+  val latestReleaseDate: LocalDate?
+    get() = if (kind == LicenceKind.PRRD) postRecallReleaseDate else conditionalReleaseDate
```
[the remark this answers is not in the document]
SebNegahbanMoJ: There aren't many other places (yet) that we need to interchangeably use CRD or PRRD depending on context. The hard stop calculation (this PR) is one of them, and the other is the early release calculation, which will need to come in future

PR #834, opened 2025-07-21
Updated staff record code to improved logic,.. also improved test cov…

AledWEvans:
…erage


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/StaffRepository.kt
```diff
@@ -9,7 +10,19 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.Staff
...
+        SELECT s FROM CommunityOffenderManager s
+            WHERE s.staffIdentifier = :staffIdentifier OR UPPER(s.username) = UPPER(:username)
+            ORDER BY s.lastUpdatedTimestamp DESC
+    """,
+  )
+  fun findCommunityOffenderManager(
+    staffIdentifier: Long,
+    username: String,
+  ): List<CommunityOffenderManager>
+
```
AledWEvans, line 25: Could have just change to JPQL from a derived query to get the order correct. I original changed this because I was the discriminator would not work on the return type, this I'm sure used to be the case but not with the latest JPA/Spring we are ok!
Could change it to
`findByStaffIdentifierOrUsernameIgnoreCaseOrderByLastUpdatedTimestampDesc` 
if you wish,.. seems a mouth full

src/test/resources/test_data/seed-community-offender-manager-with-duplicates.sql
```diff
@@ -0,0 +1,6 @@
+insert into staff(kind, staff_identifier, username, email, first_name, last_name,last_updated_timestamp)
+values
+	('PRISON_USER', 2001, 'test-client-2', 'testClient@probation.gov.uk', 'Test', 'Client','2022-12-28 09:19:17.643'),
+    ('COMMUNITY_OFFENDER_MANAGER', 2001, 'test-client-1', 'testClient@probation.gov.uk', 'Test', 'Client','2023-12-28 09:19:17.643'),
+	('COMMUNITY_OFFENDER_MANAGER', 2002, 'TEST-CLIENT-1', 'testClient@probation.gov.uk', 'Test', 'Client','2024-12-28 09:19:17.643'),
+	('PRISON_USER', 2001, 'test-client-3', 'testClient@probation.gov.uk', 'Test', 'Client','2024-12-28 09:19:17.643');
```
AledWEvans, line 6: Did try the blow combination also:
```
insert into staff(kind, staff_identifier, username, email, first_name, last_name,last_updated_timestamp) values
	('COMMUNITY_OFFENDER_MANAGER', 2005, 'TEST-CLIENT-3', 'testClient@probation.gov.uk', 'Test', 'Client','2024-12-28 09:19:17.643'),
	('COMMUNITY_OFFENDER_MANAGER', 2006, 'TEST-CLIENT-4', 'testClient@probation.gov.uk', 'Test', 'Client','2024-12-29 09:19:17.643');
```
When doing an update request base on staff_identifier = 2005 and a username = test-client-4 it will pull down the above... in the above case it will update the last
and in this case we will get a :
Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "staff_staff_identifier_key"
becuase it will try and update 2006 to 2005 and it already exists in the other row
When doing the oppsit query When doing an update request base on staff_identifier = 2006 and a username = test-client-3 I will get this error
  Detail: Key (username)=(TEST-CLIENT-3) already exists.
Make sense to me, so if this did occour in production we would get an exception some times depending on order.  

So if this does happen what should we do,... ?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/StaffIntegrationTest.kt
```diff
@@ -11,58 +11,92 @@ import org.springframework.http.MediaType
...
 
-    with(staffRepository.findAll().first() as CommunityOffenderManager) {
-      assertThat(staffIdentifier).isEqualTo(updateCom.staffIdentifier)
-      assertThat(username).isEqualTo(updateCom.staffUsername.uppercase())
-      assertThat(email).isEqualTo(updateCom.staffEmail)
-      assertThat(firstName).isEqualTo(updateCom.firstName)
-      assertThat(lastName).isEqualTo(updateCom.lastName)
-    }
+    // Then
+    val staff = staffRepository.findById(9)
```
andrewrlee, line 46: Be good for this to be more intention revealing.. it's not very obvious that a staff record has been added if we're just checking the state of a given staff record with a specific id

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/StaffIntegrationTest.kt
```diff
@@ -11,58 +11,92 @@ import org.springframework.http.MediaType
...
 
-    with(staffRepository.findAll().first() as CommunityOffenderManager) {
-      assertThat(staffIdentifier).isEqualTo(updateCom.staffIdentifier)
-      assertThat(username).isEqualTo(updateCom.staffUsername.uppercase())
-      assertThat(email).isEqualTo(updateCom.staffEmail)
-      assertThat(firstName).isEqualTo("NEW NAME")
-      assertThat(lastName).isEqualTo(updateCom.lastName)
-    }
+    // Then
+    val staff = staffRepository.findById(1)
```
andrewrlee, line 66: similarly here

andrewrlee (approved):
LGTM

PR #835, opened 2025-07-22
CVSL-3127 handling cases with null created dates

andrewrlee:
There's around 250ish of these old licences without a created date.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/publicApi/model/licence/Licence.kt
```diff
@@ -63,9 +66,12 @@ data class Licence(
...
   val createdByUsername: String,
 
-  @Schema(description = "The date and time when the version/variation of the licence was created at", example = "2023-11-20T00:00:00Z")
+  @Schema(
+    description = "The date and time when the version/variation of the licence was created at",
+    example = "2023-11-20T00:00:00Z",
+  )
   @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss")
-  val createdDateTime: LocalDateTime,
+  val createdDateTime: LocalDateTime?,
```
andrewrlee, line 74: made created date nullable

PR #849, opened 2025-07-24
CVSL:2862: PublicLicenceService updated to pull images from correct location

ryantk:
I have found another place where the fullSizeImage field was being called directly and updated to use the document service logic.


andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #859, opened 2025-07-29
CVSL-2862: Remove unused delete exclusion map endpoint

ryantk:
This endpoint is not being used by the front end


AledWEvans (approved):
LGTM

PR #863, opened 2025-07-29
CVLS-3153: Documents are deleted in licence lifecycle

ryantk:
Documents are deleted when a licence is discarded and when any additional conditions are removed.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/conditions/ExclusionZoneService.kt
```diff
@@ -82,66 +80,85 @@ class ExclusionZoneService(
...
-    }
+    // Remove AdditionalConditionUploadSummary documents
+    uploadSummaries.forEach { uuidsToDelete.add(it.thumbnailImageDsUuid) }
+
+    // Remove AdditionalConditionUploadDetail documents
+    uploadSummaries
+      .map { additionalConditionUploadDetailRepository.findById(it.uploadDetailId).orElseThrow() }
+      .forEach { uuidsToDelete.addAll(listOf(it.fullSizeImageDsUuid, it.originalDataDsUuid)) }
+
+    uuidsToDelete.filterNotNull()
```
andrewrlee, line 124: we need to make sure that the document UUIDs aren't still being used by another existing licence.
 
e.g:
* Create a licence with exclusion zone
* Vary the licence, creates a copy
* remove the exclusion zone condition from the variation
* exclusion zone still visible on the original licence pdf

ryantk: gotcha, I will take a look at this now

ryantk: The solution I have opted for is to count the number of times a specific UUID exists in the DB globally and if it appears only once then it is safe to delete. See DocumentCountsRepository.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/conditions/ExclusionZoneService.kt
```diff
@@ -148,6 +109,60 @@ class ExclusionZoneService(
       }
   }
 
+  fun deleteDocumentsFor(licence: Licence) {
```
andrewrlee, line 112: missing `@Transactional` annotations on these public functions

andrewrlee (approved):
LGTM

PR #866, opened 2025-07-30
CVSL-3156 moving preprod jobs so they don't run out of hours

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/Views.kt
justin-gardiner, line 1: was this meant to be included, seems unrelated to jobs?

justin-gardiner (approved):
LGTM

PR #868, opened 2025-07-30
CVSL-3146: Throw exception if time out is called for a licence type t…

justin-gardiner:
…hat doesn't support it.


andrewrlee (approved):
LGTM

PR #881, opened 2025-08-01
Cvsl 3154 postcode lookup get address preferred api

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/StaffController.kt
```diff
@@ -205,4 +207,46 @@ class StaffController(
...
+          Content(
+            mediaType = "application/json",
+            array = ArraySchema(schema = Schema(implementation = AddressResponse::class)),
+          ),
+        ],
+      ),
+      ApiResponse(
+        responseCode = "400",
+        description = "Bad request, staff identifier must be valid",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
```
AledWEvans, line 237: I guess your not really providing the staff ident.. any more do not sure if you need a 400 at all as there is no path or url prarams!

nishanthmahasamudram: Ya removed 400.

PR #889, opened 2025-08-05
CVSL-3036 copying image type through to SAR

andrewrlee:
Also setting image type correctly when persisting entity

Also fixing warnings


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/request/AddAddressRequest.kt
```diff
@@ -50,7 +50,7 @@ data class AddAddressRequest(
   @field:Schema(example = "MANUAL", description = "Source of the address", required = true)
   val source: AddressSource,
 
-  @Schema(
+  @field:Schema(
```
AledWEvans, line 53: why the @field?

andrewrlee: it should be param really - I was following the convention of the rest of the class

PR #890, opened 2025-08-06
CVSL-3137 Making NOT_STARTED hard stop cases redirect to PRISON_WILL_CREATE_THIS_LICENCE

SebNegahbanMoJ:
licenceId is null when the case is `NOT_STARTED`, as the "licence" being referenced is a pre-creation placeholder.


AledWEvans (approved):
LGTM

PR #891, opened 2025-08-06
CVSL-3036 move attachements response up to the same level as content

gurps-singh:
Now we are sending attachments as part of our SAR response, this PR is to move the attachments response up a level in the JSON payload.


AledWEvans (approved):
LGTM

PR #893, opened 2025-08-07
Cvsl 3193 hibernate update hibernate id generation incompatibility with kotlin default value

AledWEvans:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/VariationLicenceTest.kt
```diff
@@ -55,8 +56,10 @@ class VariationLicenceTest {
 
     VariationLicence::class.memberProperties.forEach {
       // HDCAD will always be null for non-HDC licences
-      if (it.name != "homeDetentionCurfewActualDate" && it.get(variationLicence) == null) {
-        fail { "${it.name} does not have a value set - needs to be set to test copy" }
+      if (it.name != "homeDetentionCurfewActualDate") {
+        it.isAccessible = true
```
AledWEvans, line 60: was to prevent an error occurring because of the idInternal being protected.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalConditionUploadSummary.kt
```diff
@@ -9,15 +9,16 @@ import jakarta.persistence.JoinColumn
...
 import java.time.LocalDateTime
 
 @Entity
 @Table(name = "additional_condition_upload_summary")
 data class AdditionalConditionUploadSummary(
   @Id
   @GeneratedValue(strategy = GenerationType.IDENTITY)
-  @param:NotNull
-  val id: Long = -1,
+  @param:Positive
```
andrewrlee, line 20: do these do anything?

AledWEvans: The theory is that it prevents negatives,.. but I have a feeling they are not working as exspected

PR #894, opened 2025-08-08
CVSL-3036 Adding image size column

andrewrlee:
Previously we were setting the original file size rather than the extracted image size


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/AdditionalConditionUploadSummary.kt
```diff
@@ -27,6 +27,9 @@ data class AdditionalConditionUploadSummary(
   @field:Schema(description = "The original file size in bytes", example = "27566")
   val fileSize: Int = 0,
 
+  @field:Schema(description = "The image size in bytes", example = "27566")
+  val imageSize: Int? = 0,
```
andrewrlee, line 31: currently nullable - we should be able to make this not null after running a migration

PR #898, opened 2025-08-12
CVLS-3022: Unschedule document migration job

ryantk:
The migration has finished on both pre-prod and prod so the job just runs, finds nothing and ends.
This removes the job from cron as it doesn't need to run on a schedule anymore.


AledWEvans (approved):
LGTM

PR #901, opened 2025-08-13
CVSL-3195 Hibernate update - Remove Entity Kotlin copy and update ent…

AledWEvans:
update values directly for updatedAdditionalCondition

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AbstractIdEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AbstractIdEntity.kt
@@ -23,7 +23,7 @@ abstract class AbstractIdEntity(
   @GeneratedValue(strategy = IDENTITY)
   @Column(name = "id")
   @field:Positive
-  protected val idInternal: Long? = null,
+  protected open val idInternal: Long? = null,
 ) {
 
   open val id: Long
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalCondition.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalCondition.kt
@@ -27,7 +27,7 @@ data class AdditionalCondition(
   @JoinColumn(name = "licence_id", nullable = false)
   var licence: Licence,
 
-  val conditionVersion: String,
+  var conditionVersion: String,
   val conditionCode: String,
   var conditionCategory: String,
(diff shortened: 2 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AbstractIdEntity.kt
```diff
@@ -23,7 +23,7 @@ abstract class AbstractIdEntity(
   @GeneratedValue(strategy = IDENTITY)
   @Column(name = "id")
   @field:Positive
-  protected val idInternal: Long? = null,
+  protected open val idInternal: Long? = null,
 ) {
```
AledWEvans, line 27: Some how this got lost in a merge last week

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/AdditionalCondition.kt
```diff
@@ -43,7 +43,7 @@ data class AdditionalCondition(
   )
   @Fetch(FetchMode.SUBSELECT)
   @OrderBy("dataSequence")
-  val additionalConditionData: List<AdditionalConditionData> = emptyList(),
+  val additionalConditionData: MutableList<AdditionalConditionData> = mutableListOf(),
```
AledWEvans, line 47: These values were never immutable by design,.. in the old code we used copy to add the the condition data, 

```
    val updatedAdditionalCondition = additionalCondition.copy(
      conditionVersion = version,
      additionalConditionData = additionalConditionData,
```
now we have removed copy as a method to updating the entity we must change the list to mutable its true state.

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/conditions/LicenceConditionServiceTest.kt
```diff
@@ -661,108 +661,43 @@ class LicenceConditionServiceTest {
...
-      // Verify last contact info is recorded
-      assertThat(licenceCaptor.value)
-        .extracting("updatedByUsername", "updatedBy")
-        .isEqualTo(listOf(aCom.username, aCom))
-
-      verify(conditionFormatter).format(CONDITION_CONFIG, conditionCaptor.value.additionalConditionData)
-    }
-
-    @Test
-    fun `updating additional condition data triggers checking condition formatting`() {
```
AledWEvans, line 723: this test was a duplicate of the above

justin-gardiner (approved):
LGTM

PR #905, opened 2025-08-19
CVSL-3215 Extend DB refresh timeout

andrewrlee:
This should only be required temporarily whilst we have a load of PDFs in the DB. Once they are removed we should be able to reduce this


src/main/resources/migration/adhoc/image_size_migration.sql
```diff
@@ -0,0 +1,13 @@
...
+where
+    acus.upload_detail_id = acud.id
+  and
+    acus.id in (
+    SELECT id
+    FROM   additional_condition_upload_summary
+    WHERE  image_size is null
+    LIMIT  1
+    );
+
```
andrewrlee, line 13: this has been run but want to include in source control as a reference

louisenorris (approved):
LGTM

PR #915, opened 2025-08-27
CVSL-3218: Add endpoint to synchronise COM allocation with Delius.

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/ComAllocatedHandler.kt
```diff
@@ -110,6 +110,14 @@ class ComAllocatedHandler(
     }
   }
 
+  fun syncComAllocation(crn: String) {
+    log.info("Syncing COM allocation for CRN: {}", crn)
+    val offenderManager = getOffenderManagerForDeliusEvent(crn)
```
andrewrlee, line 115: Is there a name we could give this which doesn't refer to event? `getOffenderManagerForDeliusEvent`

justin-gardiner: yes good point, I really didn't like calling ComAllocatedHandler, would prefer this code was in a common service callable by the event handler and endpoint but appeared to be a bit of work. Will update

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/ComAllocatedHandler.kt
```diff
@@ -92,13 +92,13 @@ class ComAllocatedHandler(
   private fun getOffenderManager(crn: String, event: HMPPSDomainEvent): OffenderManager? {
     val isDeliusEvent = event.detailUrl.isNullOrBlank()
     return if (isDeliusEvent) {
-      getOffenderManagerForDeliusEvent(crn)
+      getOffenderManagerFromDelius(crn)
     } else {
-      getOffenderManagerForApop(event.detailUrl!!)
```
justin-gardiner, line 97: non null assertion was unnecessary because we check event.detailUrl.isNullOrBlank() above

PR #927, opened 2025-09-04
CVSL-3176 remove the COM as being mandatory on the Licence and move t…

gurps-singh:
This PR is to make the COM optional on a licence. It removes the COM as being mandatory on the abstract Licence and instead, shifts this mandate to each individual licence kind to decide whether the responsible COM should be required or not. As part of this, the licence specification used in the `/match` endpoint has been changed as we now need to account for the different licence kinds when querying for the responsible COM in our database and the COM no longer exists on the base class.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
```diff
@@ -22,6 +28,16 @@ data class LicenceQueryObject(
   val sortOrder: String? = null,
 )
```
gurps-singh, line 30: **note:** ‏Eager load `responsibleCom` when present and use LEFT join to keep Licence rows even if they don't have a `responsibleCom` to avoid n+1. Use treat to fetch `responsibleCom` from the subclass where it is defined. Using query.distinct() to avoid duplicates produced by multiple joins.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
```diff
@@ -30,8 +46,12 @@ fun LicenceQueryObject.toSpecification(): Specification<Licence> = and(
   hasPdusIn(pdus),
   hasProbationAreaCodeIn(probationAreaCodes),
 )
-  .and { root, query, _ ->
-    root.fetch<Licence, CommunityOffenderManager>("responsibleCom", JoinType.INNER)
```
gurps-singh, line 34: **note:** ‏Eager load `responsibleCom` when present

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
```diff
@@ -30,8 +46,12 @@ fun LicenceQueryObject.toSpecification(): Specification<Licence> = and(
...
+  .and { root, query, criteriaBuilder ->
+
+    for (licenceKind in LICENCE_KINDS_WITH_RESPONSIBLE_COM) {
+      criteriaBuilder.treat(root, licenceKind).fetch<Licence, CommunityOffenderManager>("responsibleCom", JoinType.LEFT)
+    }
+    query.distinct(true)
```
gurps-singh, line 54: **note:** ‏This was suggested by CoPilot as query.distinct(true) then removes duplicates created by the fetch joins which makes sense because of the multiple joins.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
```diff
@@ -60,9 +80,19 @@ fun hasNomsIdIn(nomsIds: List<String>?): Specification<Licence>? = nomsIds?.let
...
 fun hasResponsibleComIn(staffIds: List<Int>?): Specification<Licence>? = staffIds?.let {
-  return Specification<Licence> { root, _, builder ->
-    val joinOffenderManager: Join<Licence, CommunityOffenderManager> = root.join("responsibleCom", JoinType.INNER)
-    builder.and(joinOffenderManager.get<Int>("staffIdentifier").`in`(it))
+  return Specification<Licence> { root, query, criteriaBuilder ->
+    if (staffIds.isEmpty()) return@Specification null
+
+    query.distinct(true)
+
+    val predicates = LICENCE_KINDS_WITH_RESPONSIBLE_COM.map { licenceKind ->
```
gurps-singh, line 88: ‏It builds an OR over several subclass joins. If those were INNER, the joins occur in the FROM clause and would require matches in all joined tables before the OR predicates are even evaluated. Using LEFT means non-matching joins return NULLs, so the row can still pass if any subclass predicate matches.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/CrdLicence.kt
```diff
@@ -95,6 +94,10 @@ class CrdLicence(
   )
   override var electronicMonitoringProvider: ElectronicMonitoringProvider? = null,
   override var versionOfId: Long? = null,
+
+  @ManyToOne(fetch = FetchType.EAGER)
```
gurps-singh, line 98: note: ‏Require a COM at creation and never allow setting to null

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
```diff
@@ -30,8 +46,12 @@ fun LicenceQueryObject.toSpecification(): Specification<Licence> = and(
   hasPdusIn(pdus),
   hasProbationAreaCodeIn(probationAreaCodes),
 )
-  .and { root, query, _ ->
-    root.fetch<Licence, CommunityOffenderManager>("responsibleCom", JoinType.INNER)
+  .and { root, query, criteriaBuilder ->
+
+    for (licenceKind in LICENCE_KINDS_WITH_RESPONSIBLE_COM) {
```
gurps-singh, line 51: **note**: ‏‏Here it is eager loading and fetching `responsibleCom` to avoid N+1. As the `responsibleCom` is now defined per subclass, it is now treating the root as each subclass and fetching. If any one of those joins were `INNER`, it would filter out licences that don’t belong to/ don’t have the association on that particular subclass. `LEFT` avoids that unintended filtering and keeps those rows and still pulls the association when it exists.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
gurps-singh, line 88: **note**: ‏ ‏It builds an OR over several subclass joins as `responsibleCom` is defined on subclasses, not on the base type. A single join on the base no longer works. If those were `INNER`, the joins occur in the FROM clause and would require matches in all joined tables before the OR predicates are even evaluated. Using `LEFT` means non-matching joins return NULLs, so the row can still pass if any subclass predicate matches.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/HdcLicence.kt
```diff
@@ -168,11 +172,18 @@ class HdcLicence(
...
   bespokeConditions = bespokeConditions.toMutableList(),
-  responsibleCom = responsibleCom,
   updatedBy = updatedBy,
 ),
   HasElectronicMonitoringResponseProvider,
-  HdcCase {
+  HdcCase,
+  HasCom {
+
+  @RequiresCom("Does a Hdc licence always need a COM set?")
```
andrewrlee, line 181: yes it does, you can't start to create a licence without one - it's the COM who creates it

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/VariationLicence.kt
```diff
@@ -141,10 +145,17 @@ class VariationLicence(
...
   bespokeConditions = bespokeConditions.toMutableList(),
-  responsibleCom = responsibleCom,
   updatedBy = updatedBy,
 ),
-  Variation {
+  Variation,
+  HasCom {
+
+  @RequiresCom("Do variations always need a COM set? , Does a variation licence always need a COM set?")
+  override fun getCom(): CommunityOffenderManager = responsibleCom
```
andrewrlee, line 154: Yes it does

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceFactory.kt
```diff
@@ -318,6 +327,7 @@ object LicenceFactory {
     createdBy = creator,
   )
 
+  @RequiresCom("Fired from LicenceService when creating a variation - Do we allow a HDC variation to be created where there is no COM?")
```
andrewrlee, line 330: as above - no

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -1041,46 +1059,49 @@ class LicenceService(
     )
   }
 
+  @RequiresCom("Do we timeout a time served case?, Do we timeout a licence if it has no COM allocated?")
   @Transactional
   fun timeout(licence: EntityLicence, reason: String? = null) {
     check(licence is SupportsHardStop) { "Can only timeout licence kinds that support hard stop: ${licence.id}" }
+    if (licence is HasCom) {
```
andrewrlee, line 1066: you can remove this as this is covered by the above check

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/RequiresCom.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
+
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY)
+annotation class RequiresCom(
```
andrewrlee, line 4: Can we call this `TimeServedConsideration` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -284,6 +287,7 @@ fun toVariation(
   submittedByFullName = licence.getSubmittedByFullName(),
 )
 
+@RequiresCom("Is COM required for PRRD, if so the responsibleCOM would never be null")
```
andrewrlee, line 290: As above, yes

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -570,6 +576,7 @@ fun toHdc(
   electronicMonitoringProviderStatus = determineElectronicMonitoringProviderStatus(licence.electronicMonitoringProvider),
 )
 
+@RequiresCom("Is COM required for HDC variation, if so the responsibleCOM would never be null")
```
andrewrlee, line 579: As above yes

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/ToModelTransformers.kt
```diff
@@ -862,6 +869,7 @@ fun Licence.getSubmittedByFullName(): String? {
   }
 }
 
+@RequiresCom("Without a COM, where can we get this data when setting the COM username or do we just surface results as null")
```
andrewrlee, line 872: yes, just surface null

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
```diff
@@ -22,6 +28,16 @@ data class LicenceQueryObject(
   val sortOrder: String? = null,
 )
 
+private val LICENCE_KINDS_WITH_RESPONSIBLE_COM: List<Class<out Licence>> = listOf(
```
andrewrlee, line 31: Keeping this list up to date is going to be difficult to remember and it is and always will be just all licence subtypes.. 

Think the easiest thing to do would be to add a new class field to `LicenceKind` and then build this list dynamically here: `LicenceKind.values().map{ it.clazz}`

Then we can't add a new kind without registering a class

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceSpecifications.kt
gurps-singh: Had a DM conversation with Andy around this. Works as expected re generated SQL. Not the most performant but works.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/HasCom.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity
+
+interface HasCom {
```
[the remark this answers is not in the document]
gurps-singh: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/HdcLicence.kt
```diff
@@ -230,7 +241,7 @@ class HdcLicence(
     standardConditions: List<StandardCondition> = this.standardConditions,
     additionalConditions: List<AdditionalCondition> = this.additionalConditions,
     bespokeConditions: List<BespokeCondition> = this.bespokeConditions,
-    responsibleCom: CommunityOffenderManager = this.responsibleCom,
+    responsibleCom: CommunityOffenderManager = this.responsibleCom!!,
```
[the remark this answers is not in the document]
gurps-singh: Yup, you're right - was flagged by IntelliJ. Removed

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/VariationLicence.kt
gurps-singh: Will leave this in and rejig what is written.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
gurps-singh: Removing this causes an issue when we call `sendEditedLicenceTimedOutEmail` to send the email as it can't find the COM on the entity Licence. Fixed by moving the check to that specific location only.

PR #928, opened 2025-09-04
CVSL-3287 Allowing PRRD licences to be created from the COM create caseload

SebNegahbanMoJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/LicenceController.kt
```diff
@@ -109,10 +107,9 @@ class LicenceController(
     @RequestBody @Valid
     request: CreateLicenceRequest,
   ): LicenceCreationResponse = when (request.type) {
-    PRRD -> licenceCreationService.createPrrdLicence(request.nomsId)
-    CRD -> licenceCreationService.createCrdLicence(request.nomsId)
     HARD_STOP -> licenceCreationService.createHardStopLicence(request.nomsId)
     HDC -> licenceCreationService.createHdcLicence(request.nomsId)
+    else -> licenceCreationService.createLicence(request.nomsId)
```
andrewrlee, line 112: can we mark the other two LicenceTypes as being deprecated on `CreateLicenceRequest`?

Can't remove yet as it'd break the frontend

PR #929, opened 2025-09-04
Fixing cron expression for licence activation job in preprod

andrewrlee:


SebNegahbanMoJ (changes requested):
Should this be`0 9,12,15,18 * * *`?

AledWEvans:
> Should this be`0 9,12,15,18 * * *`?

I wondered about this, but assumed it would default to 0 mins

SebNegahbanMoJ (approved):

PR #934, opened 2025-09-09
CVSL-3118 CVL Second telephone number for the initial appointment

AledWEvans:
So the rector is to move all the appointment details into a new table as well as adding a new field for the second telephone number.

**Please check if I missed out any appointment details!?**

Used the following to test migration on preprod (migration was in memory only changed the create tables to TEMP ones and removed the drop column table sql), it took 1.8 seconds.

Execute time	1.873s
Start time	Wed Sep 10 13:22:16 BST 2025
Finish time	Wed Sep 10 13:22:18 BST 2025

```
select (select count(*) from licence_appointment la), (select count(*) from appointment_address la),(select count(*) from address);

select 
	l.id,
	l.appointment_person_type,
    a.person_type,
	l.appointment_person,
	a.person,
	l.appointment_time_type,
	a.time_type,
	l.appointment_time,
	a.time,
	l.appointment_address,
	a.address_text,
	ad.first_line, 
	l.appointment_contact,
	a.telephone_contact_number,
	l.date_created,
	a.date_created,
	l.date_last_updated,
	a.date_last_updated
from licence l 
     join licence_appointment la on la.licence_id  = l.id 
     join appointment a on a.id  = la.appointment_id 
     left join appointment_address aa on aa.appointment_id  = a.id 
     left join address ad on ad.id  = aa.address_id;
```
The migration is in the fly way script


src/main/resources/migration/common/V55__create_appointment_table.sql
```diff
@@ -0,0 +1,142 @@
...
+  NEW.date_last_updated := CURRENT_TIMESTAMP;
+RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+-- 14. Trigger to update date_last_updated before any update
+CREATE TRIGGER set_appointment_last_updated_timestamp
+	BEFORE UPDATE ON appointment
+	FOR EACH ROW
+	EXECUTE FUNCTION appointment_update_last_updated_timestamp();
```
AledWEvans, line 130: This should be okay in a flyway script as it's a relatively simple sql and should be fast, I will test it on preprod!

PR #955, opened 2025-09-23
CVSL-3336 Recalls Incorrect Licence Start Date Calculation for HARD_S…

AledWEvans:
Hi Guys have a look at this see if this is what you expect?

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/TimeServedCasesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/TimeServedCasesController.kt
@@ -1,0 +1,76 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.resource.privateApi
+
+import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.Parameter
+import io.swagger.v3.oas.annotations.media.ArraySchema
+import io.swagger.v3.oas.annotations.media.Content
+import io.swagger.v3.oas.annotations.media.Schema
+import io.swagger.v3.oas.annotations.responses.ApiResponse
+import io.swagger.v3.oas.annotations.responses.ApiResponses
+import io.swagger.v3.oas.annotations.security.SecurityRequirement
+import io.swagger.v3.oas.annotations.tags.Tag
+import org.springframework.http.MediaType
+import org.springframework.security.access.prepost.PreAuthorize
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestMapping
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/util/LicenceKindFunctions.kt
```diff
@@ -3,8 +3,11 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util
 import java.time.LocalDate
 
 @TimeServedConsiderations("Any consideration here for time served licences when determining kind")
-fun determineReleaseDateKind(prrd: LocalDate?, crd: LocalDate?): LicenceKind = when {
-  // hdcad?.isTodayOrInTheFuture() == true -> LicenceKind.HDC
```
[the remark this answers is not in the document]
AledWEvans: in that case will revert this refactor

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/dates/ReleaseDateService.kt
```diff
@@ -106,15 +106,20 @@ class ReleaseDateService(
...
-    else -> calculateCrdLicenceStartDate(nomisRecord, iS91DeterminationService.isIS91Case(nomisRecord))
+
+    return when (lsdKind) {
+      HDC -> nomisRecord.homeDetentionCurfewActualDate
+      PRRD -> nomisRecord.calculatePrrdLicenceStartDate()
+      else -> calculateCrdLicenceStartDate(
+        nomisRecord,
+        iS91DeterminationService.isIS91Case(nomisRecord),
+      )
+    }
```
SebNegahbanMoJ, line 122: I personally disagree that this refactor is necessary, maybe one to chat about as a team to find the style that we prefer? If majority prefer having the assignment separately, I'm happy to move that way in future, but my view is that it's neater without

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/UpdateSentenceDatesIntegrationTest.kt
```diff
@@ -195,6 +204,80 @@ class UpdateSentenceDatesIntegrationTest : IntegrationTestBase() {
     assertThat(result?.postRecallReleaseDate).isEqualTo(postRecallReleaseDate)
   }
 
+  fun updateHardStopDateScenarios(): List<Arguments> {
+    val workingDays = workingDaysService.workingDaysAfter(LocalDate.now())
```
SebNegahbanMoJ, line 208: This is making a real API call, right?

AledWEvans: well yes but to a stubbed response

PR #968, opened 2025-09-29
CVSL-3197: Add a get licence permissions endpoint to check access a u…

justin-gardiner:
…ser has to a licence.


louisenorris (approved):
LGTM

PR #969, opened 2025-09-30
CVSL-3244 Add delius staff code to COM

andrewrlee:


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AppointmentServiceTest.kt
```diff
@@ -251,95 +248,10 @@ class AppointmentServiceTest {
...
+    val aCom = communityOffenderManager()
 
-    val aPreviousUser = CommunityOffenderManager(
-      staffIdentifier = 4000,
-      username = "test",
-      email = "test@test.com",
-      firstName = "Test",
-      lastName = "Test",
-    )
+    val aPreviousUser = communityOffenderManager()
```
SebNegahbanMoJ, line 255: **suggestion:** I think setting `aCom` and `aPreviousUser` to the same value makes the test that compares the two slightly harder to track.
I think the test still demonstrates the code working, but a bit less intuitively. Could we change one (any) of the fields to demonstrate that `aCom` != `aPreviousUser`?

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/TestData.kt
```diff
@@ -163,20 +174,8 @@ object TestData {
...
-    ),
-    createdBy = CommunityOffenderManager(
-      staffIdentifier = 2000,
-      username = "tcom",
-      email = "testemail@probation.gov.uk",
-      firstName = "X",
-      lastName = "Y",
-    ),
+    responsibleCom = uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.communityOffenderManager(),
+    createdBy = uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.communityOffenderManager(),
```
SebNegahbanMoJ, line 178: Can we import these under aliases if the name is shared?

louisenorris (approved):
LGTM

PR #978, opened 2025-10-03
CVSL-3316 Adding CvlRecordService

SebNegahbanMoJ:
This is an obviously monstrous PR. To hopefully ease the pain, I've tried to divide the changes into sensible commit chunks, grouping sub-functions together, and I've left some comments to hopefully explain some bits. The side-effect of the grouping however is that some places may include references to things that don't exist in that commit, but are introduced later. If anything seems to randomly appear unexplained, hopefully checking the final result of the file will make sense.

I'm happy to be on as many calls as needed to explain bits if anything isn't clear.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
@@ -16,8 +16,8 @@ class EligibilityService(
 ) {
 
   fun getEligibilityAssessment(prisoner: PrisonerSearchPrisoner, areaCode: String): EligibilityAssessment {
-    val assessment = getEligibilityAssessments(listOf(prisoner), mapOf(prisoner.prisonerNumber to areaCode))
-    return assessment[prisoner.prisonerNumber] ?: error("Unable to find eligibility assessment for ${prisoner.prisonerNumber}")
+    val assessments = getEligibilityAssessments(listOf(prisoner), mapOf(prisoner.prisonerNumber to areaCode))
+    return assessments.values.first()
   }
 
   fun getEligibilityAssessments(


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CvlRecordService.kt
```diff
@@ -0,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.dates.ReleaseDateService
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonerSearchPrisoner
+
+@Service
+class CvlRecordService(
```
SebNegahbanMoJ, line 8: This class is the crux of the changes. In almost all cases, the eligibility service is no longer called directly, nor is `getLicenceStartDate(s)`. The only places that call the eligibility service directly are those where licences already exist, meaning LicenceStartDate does not need to be recalculated.
By utilising this with the recently eligibility service refactor, we no longer need to have `determineReleaseDateKind`, as it is replaced by `eligibleKind`. This in turn means we no longer have two places where eligibility is calculated (for `NOT_STARTED` cases via the eligibility service, and on-creation via the licence creation service)

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CvlRecordService.kt
```diff
@@ -0,0 +1,34 @@
...
+  private val eligibilityService: EligibilityService,
+  private val releaseDateService: ReleaseDateService,
+) {
+
+  fun getCvlRecord(prisoner: PrisonerSearchPrisoner, areaCode: String): CvlCaseDto {
+    val recordList = getCvlRecords(listOf(prisoner), mapOf(prisoner.prisonerNumber to areaCode))
+    return recordList.find { it.nomisId == prisoner.prisonerNumber }!!
```
SebNegahbanMoJ, line 15: This pattern of `find{...}!!` is used in a few places in regard to CvlRecords. There is no filtering of cases during either the CvlRecord building or the eligibility checking processes, and both are run based on PrisonerSearchPrisoners. This means that any case going in should have a record coming out, and it will be keyed by NOMIS ID.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/PrisonerWithCvlFields.kt
```diff
@@ -42,13 +42,3 @@ data class PrisonerWithCvlFields(
...
 )
-
-@Schema(description = "A combination of the NOMIS prisoner record and their respective licence start date")
-data class CaseloadItem(
-  @field:Schema(description = "Details about a prisoner")
-  val prisoner: Prisoner,
-
-  @field:Schema(description = "Date that the licence is due to activate", example = "05/05/2023")
-  @field:JsonFormat(pattern = "dd/MM/yyyy")
-  val licenceStartDate: LocalDate? = null,
```
SebNegahbanMoJ, line 53: Licence Start Date now comes from the CvlRecord, no need to fetch it as a part of a CaseloadService call, so this data class is redundant.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -225,12 +223,12 @@ class LicenceCreationService(
   }
 
   @TimeServedConsiderations("Any special logic for time served cases needed here - seems this is just for hard stop?")
-  fun determineLicenceKind(nomisRecord: PrisonerSearchPrisoner, licenceStartDate: LocalDate?): LicenceKind {
+  fun determineLicenceKind(cvlRecord: CvlCaseDto): LicenceKind {
     val today = LocalDate.now()
 
-    var kind = determineReleaseDateKind(nomisRecord.postRecallReleaseDate, nomisRecord.conditionalReleaseDate)
+    var kind = cvlRecord.eligibleKind ?: LicenceKind.CRD
```
SebNegahbanMoJ, line 227: Defaulting to CRD to match old behaviour. Perhaps throwing an error would be more suitable here, happy to discuss.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadSearchService.kt
```diff
@@ -69,16 +71,11 @@ class ComCaseloadSearchService(
...
     }
 
-    val prisonersWithoutLicences = cvlSearchRecords.filter { searchRecord -> searchRecord.licence == null }.mapNotNull {
-      it.prisonerSearchPrisoner
-    }
-    val licenceStartDates = releaseDateService.getLicenceStartDates(prisonersWithoutLicences)
-
     val searchResults = cvlSearchRecords.mapNotNull {
-      val licenceStartDate = licenceStartDates[it.prisonerSearchPrisoner?.prisonerNumber]
+      val cvlRecord = cvlRecords.find { cvlRecord -> it.prisonerSearchPrisoner?.prisonerNumber == cvlRecord.nomisId }
```
SebNegahbanMoJ, line 75: `CvlRecord` is being handled as nullable here, as the calling map is based on delius record, not prisoner search record. Therefore it's feasible that NOMIS ID could be null, and we wouldn't want to blow up here for that, but rather filter them out.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/dates/ReleaseDateService.kt
```diff
@@ -89,31 +89,26 @@ class ReleaseDateService(
...
-
-    return when (lsdKind) {
-      HDC -> nomisRecord.homeDetentionCurfewActualDate
-      PRRD -> nomisRecord.calculatePrrdLicenceStartDate()
-      else -> calculateCrdLicenceStartDate(
-        nomisRecord,
-        iS91DeterminationService.isIS91Case(nomisRecord),
-      )
-    }
+    else -> null
```
SebNegahbanMoJ, line 100: A null `eligibleKind` (as a result of being made ineligible for CVL) now means LSD will be nulled, making the case `attention needed`. This seemed better than defaulting to using CRD dates.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -1124,18 +1121,14 @@ class LicenceService(
...
-  private fun assertCaseIsEligible(licence: EntityLicence, nomisId: String?) {
-    if (nomisId == null) {
-      throw ValidationException("Unable to perform action, licence ${licence.id} is missing NOMS ID")
-    }
-
-    val prisoner = prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(nomisId)).first()
-    if (!eligibilityService.isEligibleForCvl(prisoner, licence.probationAreaCode)) {
-      throw ValidationException("Unable to perform action, licence ${licence.id} is ineligible for CVL")
+  private fun assertCaseIsEligible(eligibilityAssessment: EligibilityAssessment, licenceId: Long) {
+    println(eligibilityAssessment)
```
SebNegahbanMoJ: Removed in later commit

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CvlRecordService.kt
andrewrlee: here you could just use `first()`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CvlCaseDto.kt
```diff
@@ -0,0 +1,24 @@
...
+)
+
+data class EligibilityAssessment(
+  val genericIneligibilityReasons: List<String> = emptyList(),
+  val crdIneligibilityReasons: List<String> = emptyList(),
+  val prrdIneligibilityReasons: List<String> = emptyList(),
+  val isEligible: Boolean,
+  val eligibleKind: LicenceKind? = when (isEligible) {
+    true if crdIneligibilityReasons.isEmpty() -> LicenceKind.CRD
+    true if prrdIneligibilityReasons.isEmpty() -> LicenceKind.PRRD
```
andrewrlee, line 21: Get a warning on these: 
```The feature "when guards" is experimental and should be enabled explicitly. This can be done by supplying the compiler argument '-Xwhen-guards', but note that no stability guarantees are provided```

File: src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/dates/LsdRecalculationService.kt [hunk 9]
```diff
@@ -32,7 +32,11 @@ class LsdRecalculationService(
     }
 
     val prisoners = this.prisonerSearchApiClient.searchPrisonersByNomisIds(prisonNumbers)
-    val prisonersToLSDs = releaseDateService.getLicenceStartDates(prisoners)
+    val nomisIdsToKinds = prisoners.associate { prisoner ->
+      val licence = licences.find { licence -> prisoner.prisonerNumber == licence.nomsId }!!
+      return@associate prisoner.prisonerNumber to licence.kind
+    }
+    val prisonersToLSDs = releaseDateService.getLicenceStartDates(prisoners, nomisIdsToKinds)
```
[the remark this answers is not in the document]
andrewrlee: We should look at removing this as we have no plans to run it in the future

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/CaseloadController.kt
```diff
@@ -72,7 +72,7 @@ class CaseloadController(
         content = [
           Content(
             mediaType = "application/json",
-            array = ArraySchema(schema = Schema(implementation = CaseloadItem::class)),
```
andrewrlee, line 75: This function will be able to go soon once we enable to approver caseload

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/support/SupportService.kt
```diff
@@ -6,21 +6,25 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.HdcService
...
 ) {
   fun getIneligibilityReasons(prisonNumber: String): List<String> {
     val prisoners = prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(prisonNumber))
     if (prisoners.size != 1) {
       error("Found ${prisoners.size} prisoners for: $prisonNumber")
     }
     val prisoner = prisoners.first()
-    val reasons = eligibilityService.getIneligibilityReasons(prisoner)
+    val deliusRecord = deliusApiClient.getOffenderManagers(listOf(prisoner.prisonerNumber)).first()
+    val eligibility = eligibilityService.getEligibilityAssessments(listOf(prisoner), mapOf(prisoner.prisonerNumber to deliusRecord.team.provider.code))[prisoner.prisonerNumber]!!
```
andrewrlee, line 26: can use `first` here...

andrewrlee: extracting out `mapOf(prisoner.prisonerNumber to deliusRecord.team.provider.code)` might be nice too

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CvlRecordServiceTest.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    val result = service.getCvlRecord(aPrisonerSearchPrisoner, "AREA1")
+
+    assertThat(result).isEqualTo(
+      CvlCaseDto(
+        nomisId = aPrisonerSearchPrisoner.prisonerNumber,
+        licenceStartDate = LocalDate.of(2021, 10, 22),
+        isEligible = true,
+        eligibleKind = LicenceKind.CRD,
+        ineligiblityReasons = anEligibilityAssessment(),
```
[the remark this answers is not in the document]
andrewrlee: which would make tests a little bit simpler - default to empty string, don't need to specify if it's an eligible case

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -276,18 +273,24 @@ class CaCaseloadService(
...
+    val nomisIdsToAreaCodes = nomisRecordsToDeliusRecords.map { (nomisRecord, deliusRecord) -> nomisRecord.prisonerNumber to deliusRecord.team.provider.code }.toMap()
 
-    val eligibleCases = filterOffendersEligibleForLicence(casesWithoutLicences)
+    val cvlRecords = cvlRecordService.getCvlRecords(prisonersWithoutLicences, nomisIdsToAreaCodes)
 
-    return createNotStartedLicenceForCase(eligibleCases)
+    val casesWithoutLicences = buildManagedCaseDto(nomisRecordsToDeliusRecords, cvlRecords)
+
+    val eligibleCases = filterOffendersEligibleForLicence(casesWithoutLicences, cvlRecords)
+
```
andrewrlee, line 284: **suggestion:** rather than passing the full list of CVL records, you could build here of prison number to kind and pass that in to  `createNotStartedLicenceForCase`

`val prisonNumberToKind = cvlRecords.associate { it.nomisId to it.eligibleKind }`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -395,41 +391,42 @@ class CaCaseloadService(
...
   }
 
-  private fun pairNomisRecordsWithDelius(
-    prisoners: List<PrisonerSearchPrisoner>,
-    licenceStartDates: Map<String, LocalDate?>,
-  ): List<ManagedCaseDto> {
-    val caseloadNomisIds = prisoners
-      .map { it.prisonerNumber }
-
+  private fun pairNomisRecordsWithDelius(nomisRecords: List<PrisonerSearchPrisoner>): Map<PrisonerSearchPrisoner, CommunityManager> {
```
andrewrlee, line 394: **suggestion:** We're not really using the return type of this as a map, it might make more sense to return this as a list of pairs. 

It means you can simplify `currentMap.map{}.toMap()` to `newList.associate{}` in one place.

Maps with complex keys are a bit gnarly to debug and can be a source of bugs - as these are (currently) an immutable data class then it's less of a concern but it would simplify the code slightly and not have significant downsides

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
```diff
@@ -395,41 +391,42 @@ class CaCaseloadService(
...
       }
+    }.toMap()
   }
 
+  private fun buildManagedCaseDto(
+    nomisRecordsToComRecords: Map<PrisonerSearchPrisoner, CommunityManager>,
+    cvlRecords: List<CvlCaseDto>,
+  ): List<ManagedCaseDto> = nomisRecordsToComRecords
+    .map { (nomisRecord, com) ->
+      val licenceStartDate = cvlRecords.first { it.nomisId == nomisRecord.prisonerNumber }.licenceStartDate
```
andrewrlee, line 412: **suggestion:** similarly here you could pass in a `Map<prison number, Licence start date>` rather than `List<CvlCaseDto>`

`val prisonNumberToLicenceStartDate = cvlRecords.associate { it.nomisId to it.licenceStartDate }`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadSearchService.kt
```diff
@@ -56,11 +55,13 @@ class ComCaseloadSearchService(
       ),
     )
 
-    val deliusRecordsWithLicences = teamCaseloadResult.content.map { it to getLicence(it.crn) }
+    val deliusRecordsToLicences = teamCaseloadResult.content.map { it to getLicence(it.crn) }
+    val prisonerRecords = findPrisonersForRelevantRecords(deliusRecordsToLicences)
 
-    val prisonerRecords = findPrisonersForRelevantRecords(deliusRecordsWithLicences)
+    val nomisIdsToAreaCodes = teamCaseloadResult.content.mapNotNull { if (it.nomisId != null) it.nomisId to it.team.provider.code else null }.toMap()
+    val cvlRecords = cvlRecordService.getCvlRecords(prisonerRecords.map { (_, nomisRecord) -> nomisRecord }, nomisIdsToAreaCodes)
```
andrewrlee, line 62: **suggestion:** `cvlRecordService.getCvlRecords(prisonerRecords.values.toList(), nomisIdsToAreaCodes)`

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadSearchService.kt
andrewrlee: I'm not generally a bit fan of specifying the type if its obvious based on context but this might be a good point to explicitly state it?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadSearchService.kt
```diff
@@ -109,36 +105,25 @@ class ComCaseloadSearchService(
...
-        determineReleaseDateKind(
-          licence.postRecallReleaseDate,
-          licence.conditionalReleaseDate,
-        ) == LicenceKind.PRRD ||
-          eligibilityService.isEligibleForCvl(
-            prisonOffender,
-            deliusOffender.team.provider.code,
-          )
-        ) ->
-      deliusOffender.toStartedRecord(licence)
```
andrewrlee, line 140: 🥳

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CaseloadService.kt
```diff
@@ -2,39 +2,29 @@ package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
...
-      )
-    }
-  }
+  fun getPrisonersByNumber(nomisIds: List<String>): List<PrisonerSearchPrisoner> = prisonerSearchApiClient.searchPrisonersByNomisIds(nomisIds)
 
   fun getPrisoner(nomisId: String): PrisonerWithCvlFields {
     val prisoner = prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(nomisId)).firstOrNull() ?: throw EntityNotFoundException(nomisId)
-    val licenceKind = determineReleaseDateKind(prisoner.postRecallReleaseDate, prisoner.conditionalReleaseDate)
-    val licenceStartDate = releaseDateService.getLicenceStartDate(prisoner, licenceKind)
+    val comRecord = deliusApiClient.getOffenderManager(nomisId)!!
```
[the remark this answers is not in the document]
justin-gardiner: Are we sure this won't return null?  Is it possible here that they haven't been assigned an offender manager? I don't know, just interested.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
justin-gardiner: I would have thought not being able to determine a kind would be an error? Maybe we should at least log it?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CvlRecordService.kt
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.prison.PrisonerSearchPrisoner
+
+@Service
+class CvlRecordService(
+  private val eligibilityService: EligibilityService,
+  private val releaseDateService: ReleaseDateService,
+) {
+
+  fun getCvlRecord(prisoner: PrisonerSearchPrisoner, areaCode: String): CvlCaseDto {
+    val recordList = getCvlRecords(listOf(prisoner), mapOf(prisoner.prisonerNumber to areaCode))
```
AledWEvans, line 14: nitpick: getCvlRecord but we return a CvlCaseDto, shoul'nt these share a common name,...

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -74,17 +73,15 @@ class LicenceCreationService(
     val deliusRecord = deliusApiClient.getProbationCase(prisonNumber)
     val prisonInformation = prisonApiClient.getPrisonInformation(nomisRecord.prisonId!!)
     val currentResponsibleOfficerDetails = getCurrentResponsibleOfficer(deliusRecord, prisonNumber)
+    val cvlRecord = cvlRecordService.getCvlRecord(nomisRecord, currentResponsibleOfficerDetails.team.provider.code)
```
AledWEvans, line 76: nickpick: is it a cvlRecord or a CvlCaseDto

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CvlCaseDto.kt
```diff
@@ -0,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service
+
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceKind
+import java.time.LocalDate
+
+data class CvlCaseDto(
```
AledWEvans, line 6: CvlCaseRacordDto ?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadSearchService.kt
```diff
@@ -56,11 +55,13 @@ class ComCaseloadSearchService(
       ),
     )
 
-    val deliusRecordsWithLicences = teamCaseloadResult.content.map { it to getLicence(it.crn) }
+    val deliusRecordsToLicences = teamCaseloadResult.content.map { it to getLicence(it.crn) }
+    val prisonerRecords = findPrisonersForRelevantRecords(deliusRecordsToLicences)
 
-    val prisonerRecords = findPrisonersForRelevantRecords(deliusRecordsWithLicences)
+    val nomisIdsToAreaCodes = teamCaseloadResult.content.mapNotNull { if (it.nomisId != null) it.nomisId to it.team.provider.code else null }.toMap()
```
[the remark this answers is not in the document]
AledWEvans: My preference would be
```
val nomisIdsToAreaCodes = teamCaseloadResult.content
    .mapNotNull { it.nomisId?.let { id -> id to it.team.provider.code } }
    .toMap()
```

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCreateCaseloadService.kt
```diff
@@ -94,28 +96,22 @@ class ComCreateCaseloadService(
...
-    }
+
     when {
-      prisoner.bookingId == null -> false
+      nomisRecord.bookingId == null -> false
       licences.any { it.licenceStatus == ACTIVE } -> false
-      prrdLicence != null && prrdLicence.licenceId != null -> true
-      !eligibilityService.isEligibleForCvl(prisoner, deliusRecord.team?.provider?.code) -> false
-      else -> true
+      else -> cvlRecord.isEligible
```
AledWEvans, line 109: nice

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/support/SupportService.kt
```diff
@@ -6,23 +6,26 @@ import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.HdcService
...
 ) {
   fun getIneligibilityReasons(prisonNumber: String): List<String> {
     val prisoners = prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(prisonNumber))
     if (prisoners.size != 1) {
       error("Found ${prisoners.size} prisoners for: $prisonNumber")
     }
     val prisoner = prisoners.first()
-    val reasons = eligibilityService.getIneligibilityReasons(prisoner)
+    val deliusRecord = deliusApiClient.getOffenderManagers(listOf(prisoner.prisonerNumber)).first()
+    val eligibility = eligibilityService.getEligibilityAssessment(prisoner, deliusRecord.team.provider.code)
```
AledWEvans, line 26: nitpick: is eligibility an eligibilityAssessment

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCaseloadSearchService.kt
SebNegahbanMoJ: I went with this approach to be explicit about the type of `values`, as I've found it a bit enigmatic in the past to refer to `key` and `value`, but happy to do it in this case as it does neaten up the param declaration

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/support/SupportService.kt
SebNegahbanMoJ: It is, I've battled with myself a little on this, as `eligibilityAssessment` is quite long to write out everywhere, but neither `eligibility` or `assessment` are especially accurate as a shorthand 😅

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
SebNegahbanMoJ: Truthfully I think the better approach here is probably just to throw an error.
It suggests the case is no longer eligible, yet we're still allowing licence creation and defaulting to CRD.
I only left this in to match existing behaviour because I felt like it may need a reasonable regression on it's own, and this is big enough!

Happy to put in the `error()` call now though if no one can think of a reason not to

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/CaCaseloadService.kt
SebNegahbanMoJ: I went a slightly different direction and added the necessary fields to ManagedCaseDto, that way I don't have to pass anything else around

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/CaseloadController.kt
```diff
@@ -72,7 +72,7 @@ class CaseloadController(
...
+            array = ArraySchema(schema = Schema(implementation = PrisonerSearchPrisoner::class)),
```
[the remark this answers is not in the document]
andrewrlee: on the surface yes.. but it's a prisoner from prisoner search so it kind of makes sense

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCreateCaseloadService.kt
```diff
@@ -55,36 +54,39 @@ class ComCreateCaseloadService(
 
   private fun buildCreateCaseload(managedOffenders: List<ManagedOffenderCrn>): List<ComCase> {
     val deliusAndNomisRecords = pairDeliusRecordsWithNomis(managedOffenders)
-    val crnsToLicenceLists = mapCasesToLicences(deliusAndNomisRecords)
-    val eligibleCases = filterCasesEligibleForCvl(deliusAndNomisRecords, crnsToLicenceLists)
+    val nomisIdsToAreaCodes = deliusAndNomisRecords.map { (deliusRecord, nomisRecord) -> nomisRecord.prisonerNumber to (deliusRecord.team?.provider?.code ?: "") }.toMap()
+    val cvlRecords = cvlRecordService.getCvlRecords(deliusAndNomisRecords.map { (_, nomisRecord) -> nomisRecord }, nomisIdsToAreaCodes)
+    val crnsToLicenceLists = mapCasesToLicences(deliusAndNomisRecords, cvlRecords)
+    val eligibleCases = filterCasesEligibleForCvl(deliusAndNomisRecords, crnsToLicenceLists, cvlRecords)
```
andrewrlee, line 60: This is now pretty dense and difficult to read, I think because it's mixing up different levels of abstraction. 

Maybe we just leave it for now and look at tidying up afterwards

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/ComCreateCaseloadService.kt
```diff
@@ -129,13 +125,13 @@ class ComCreateCaseloadService(
...
+    nomisRecord: PrisonerSearchPrisoner,
+    cvlRecord: CvlRecord,
   ): CaseLoadLicenceSummary {
-    val prisoner = nomisRecord.prisoner.toPrisonerSearchPrisoner()
-    val licenceType = LicenceType.getLicenceType(prisoner)
-    val licenceKind = licenceCreationService.determineLicenceKind(prisoner, nomisRecord.licenceStartDate)
-    val name = "${prisoner.firstName} ${prisoner.lastName}".trim().convertToTitleCase()
-    val sentenceDateHolder = prisoner.toSentenceDateHolder(nomisRecord.licenceStartDate)
+    val licenceType = LicenceType.getLicenceType(nomisRecord)
+    val licenceKind = licenceCreationService.determineLicenceKind(cvlRecord)
```
andrewrlee, line 132: **question:** why do we use the kind here rather than the one on the CvlRecord?

PR #979, opened 2025-10-06
CVSL-3351: Add licence kind to CaseloadItem to we can check if an uns…

justin-gardiner:
…tarted licence is a recall or not.


louisenorris (approved):
LGTM

PR #1335, opened 2026-04-09
CVSL-3983 Removing the old prisoner details updating logic

andrewrlee:
event processing occurs in the API now


justin-gardiner (approved):
LGTM

PR #1341, opened 2026-04-14
CVSL-3971: Process the recall inserted event by deactivating any acti…

justin-gardiner:
…ve licences for the recalled offender.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/RecallInsertedHandler.kt
```diff
@@ -0,0 +1,61 @@
...
+  companion object {
+    private val log = LoggerFactory.getLogger(PrisonerUpdatedHandler::class.java)
+  }
+
+  fun handleEvent(message: String) {
+    val event = try {
+      mapper.readValue(message, HMPPSDomainEvent::class.java)
+    } catch (e: JacksonException) {
+      log.error("Failed to parse event message", e)
+      return
```
andrewrlee: should we throw this in order to allow us to potentially process again?  is this what we do for other events?

justin-gardiner: We've got a mix of behaviour. I would have thought if the message is invalid we'd never be able to reprocess so pointless resending it.

justin-gardiner: Actually, I guess we should as it will eventually end up on the dead letter queue to be checked.

andrewrlee: it might be that the payload changes or something - we fix our marshalling class and then we're able to process the messages again

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/RecallInsertedHandler.kt
```diff
@@ -0,0 +1,67 @@
...
+  private val licenceRepository: LicenceRepository,
+  private val licenceService: LicenceService,
+  private val prisonService: PrisonService,
+  @param:Value("\${feature.toggle.standardRecalls.enabled:false}") private val standardRecallsEnabled: Boolean = false,
+) {
+  companion object {
+    private val log = LoggerFactory.getLogger(PrisonerUpdatedHandler::class.java)
+  }
+
+  fun handleEvent(message: String) {
```
andrewrlee, line 30: Do we need an `@Transactional` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/RecallInsertedHandler.kt
justin-gardiner: updated

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/RecallInsertedHandler.kt
justin-gardiner: added

andrewrlee (approved):
LGTM

PR #1343, opened 2026-04-15
CVSL-3971: Don't check for standard recalls when processing date chan…

justin-gardiner:
…ge events, it's not handled for recall inserted events.


AledWEvans (approved):
LGTM

PR #1345, opened 2026-04-16
CVSL-4000: Deactivate an active licence if an offender has a fixed te…

justin-gardiner:
…rm recall sentence added.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/RecallInsertedHandler.kt
```diff
@@ -50,12 +46,21 @@ class RecallInsertedHandler(
     val activeLicence = getActiveLicence(nomisId)
     if (activeLicence != null) {
       log.info("nomisId: $nomisId, has active licence: ${activeLicence.id}")
-      val hasStandardRecall = prisonService.hasStandardRecallSentence(nomisRecord.bookingId?.toLong()!!)
-      if (hasStandardRecall) {
+
+      val sentenceAndRecallTypes = prisonService.getSentenceAndRecallTypes(nomisRecord.bookingId?.toLong()!!)
+      val hasStandardRecall = sentenceAndRecallTypes.sentenceTypeRecallTypes.any { it.recallType.isStandardRecall }
+      val hasFixedTermRecall = sentenceAndRecallTypes.sentenceTypeRecallTypes.any { it.recallType.isFixedTermRecall }
```
andrewrlee: suggestion (non-blocking): could push this into prisonService and expose a `getRecallType` returning nullable enum. 

The benefit of doing this would be potentially simpler code here and we'd not be duplicating the prioritisation of standard recall over fix term recalls.

andrewrlee (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/RecallInsertedHandler.kt
justin-gardiner: Good idea, I'll do that.

justin-gardiner: done

andrewrlee (approved):
LGTM

PR #1348, opened 2026-04-16
CVSL-4007 add suppression file

gurps-singh:
This PR is to add a suppression file to suppress the handlebars CVEs which we are being alerted for. This is because it is coming from the subject access request library and we don't think the HAAR team are setting the property


AledWEvans (approved):
LGTM

PR #1369, opened 2026-04-23
CVSL-3991 add feature toggle for restricted patients

gurps-singh:
This PR introduces the restricted patients feature toggle to the API.

### DIFF
--- a/src/test/resources/application-test.yml
+++ b/src/test/resources/application-test.yml
@@ -116,10 +116,9 @@ feature:
   toggle:
     hdc:
       enabled: true
-    timeServed:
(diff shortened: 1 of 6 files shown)


src/test/resources/application-test.yml
```diff
@@ -116,11 +116,10 @@ feature:
   toggle:
     hdc:
       enabled: true
-    timeServed:
```
gurps-singh, line 119: **note:** ‏Seems like this was forgotten when time served flag and prisons was removed.

dave-virden (approved):
LGTM

PR #1373, opened 2026-04-24
CVSL-3971: Rename DateChangeLicenceDeactivationReason to LicenceDeact…

justin-gardiner:
…ivationReason as no longer only used for date change events.CVSL-3971: Rename DateChangeLicenceDeactivationReason to LicenceDeactivationReason as no longer only used for date change events.


AledWEvans (approved):
LGTM

PR #1377, opened 2026-04-27
CVSL-4019 Prog HDC - Prevent migrating HDC licence if there are any i…

AledWEvans:
…nflight licenses in CVL


src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/integration/migration/MigrationControllerIntegrationTest.kt
```diff
@@ -85,6 +85,48 @@ class MigrationControllerIntegrationTest : IntegrationTestBase() {
...
+    val result = webTestClient.post()
+      .uri(MIGRATE_URL)
+      .contentType(MediaType.APPLICATION_JSON)
+      .headers(setAuthorisation(roles = listOf("ROLE_CVL_ADMIN")))
+      .bodyValue(request)
+      .exchange()
+
+    // Then
+    result.expectStatus().isOk
+    assertThat(testRepository.hasMetaData()).isFalse
```
SebNegahbanMoJ, line 127: We could test the log messages here to make sure that they're being produced correctly

SebNegahbanMoJ (approved):

PR #1378, opened 2026-04-27
Cvsl 4022 cvl | cannot read exclusion zone uploads using jbig2 image

AledWEvans:
CVSL-4022 CVL | Cannot read exclusion zone uploads using JBIG2 image

Have generated a simple PDF and verifies JBIG2 compression for PDFBox testing - see ticket for full script!


andrewrlee (approved):
LGTM

PR #1381, opened 2026-04-28
CVSL-4022 CVL | Cannot read exclusion zone uploads using JBIG2 image …

AledWEvans:
update to the test pdf file!


andrewrlee (approved):
LGTM

PR #1385, opened 2026-04-30
CVSL-3980 change creation functions to work with restricted patients

gurps-singh:
This PR is to change the licence creation logic so that it works for restricted patients where they have a supporting prison ID and a prison code of "OUT". We now use supporting prison ID where they are identified as a restricted patient instead of prison code.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -65,7 +69,9 @@ class LicenceCreationService(
     val nomisRecord = prisonerSearchApiClient.searchPrisonersByNomisIds(listOf(prisonNumber)).first()
 
     val deliusRecord = deliusApiClient.getProbationCase(prisonNumber)
-    val prisonInformation = prisonApiClient.getPrisonInformation(nomisRecord.prisonId!!)
+
+    val prisonInformation = getPrisonInformation(nomisRecord)
```
gurps-singh, line 73: **note:** ‏Needed to change this as a restricted patient has a prison id of OUT and was falling over here. Now we check if they are a restricted patient and use supporting Prison ID instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceCreationService.kt
```diff
@@ -90,6 +96,7 @@ class LicenceCreationService(
         responsibleCom = responsibleCom,
         creator = createdBy,
         licenceStartDate = cvlRecord.licenceStartDate,
+        restrictedPatientsEnabled = restrictedPatientsEnabled,
```
andrewrlee: you're already passing through the prison information can you use the prison code/id off that in the factory instead instead of passing this boolean through?

gurps-singh: Thanks Andy, did not think of that - we can and should. Change incoming.

andrewrlee (approved):
LGTM

PR #1390, opened 2026-05-06
CVSL-4025: Expire AP_PSS licences that have an LED in the past.

dave-virden:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/ISRPssProgressionChunkService.kt
```diff
@@ -65,6 +71,15 @@ class ISRPssProgressionChunkService(
     log.info("ISR PSS repealed licence chunk completed. size={} batch {} to {}", licenceIds.size, licenceIds.first(), licenceIds.last())
   }
 
+  private fun createApExpiredAuditEvent(
+    licence: Licence,
+    licenceIds: List<Long>,
+  ): AuditEvent {
+    val summary = "ISR ${LicenceType.AP.name} licence for ${licence.forename} ${licence.surname} has expired as the LED is in the past"
+    val changes = mapOf("type" to summary, "changes" to mapOf("statusCode" to licence.statusCode.name))
```
andrewrlee: can we include the status code before and after? similar to `createPssAuditEvent`

PR #1391, opened 2026-05-07
CVSL-4090 HDC - CVL Migration : prevent duplicate licences being HDC …

AledWEvans:
HDC - CVL Migration : prevent duplicate licences being HDC to CVL MIGRATION committed at the same time due to race condition. Idea is it will cause a role back in one of the threads report the error back to HDC api,.. might have to find the exception type and make it none retry able..


andrewrlee (approved):
LGTM

PR #1397, opened 2026-05-11
CVL Migration: Add a try catch to intercept  DataIntegrityViolation e…

AledWEvans:
This PR is to make sure data violations on the migration table are treated as non retry able,.. We can test for this as this will only occour in a race condition between two threads


louisenorris (approved):
LGTM

PR #1398, opened 2026-05-11
ISR new status for progression

AledWEvans:
New status codes to be covered by AP_PSS and PSS migration


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/ISRProgressionLicenceRepository.kt
```diff
@@ -13,7 +13,7 @@ interface ISRProgressionLicenceRepository : JpaRepository<Licence, Long> {
     value = """
             SELECT l.id
             FROM licence l
-            WHERE l.status_code IN ('IN_PROGRESS','SUBMITTED','APPROVED','ACTIVE')
```
justin-gardiner, line 16: Is there an integration test that could be updated with one of the new status?

justin-gardiner (approved):
LGTM

PR #1399, opened 2026-05-11
CVSL-4100 Preventing PSS cases from being created

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CvlRecordService.kt
```diff
@@ -63,27 +57,10 @@ class CvlRecordService(
         isDueToBeReleasedInTheNextTwoWorkingDays = releaseDateService.isDueToBeReleasedInTheNextTwoWorkingDays(
           licenceStartDate,
         ),
-        licenceType = getLicenceType(prisoner, licenceStartDate, eligibility.eligibleKind),
+        licenceType = AP,
```
andrewrlee, line 60: All licences should be AP from now on

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -159,13 +160,10 @@ class EligibilityService(
 
   private fun hasPrrdTodayOrInTheFuture(prisoner: PrisonerSearchPrisoner): Boolean = prisoner.postRecallReleaseDate == null || dateIsTodayOrFuture(prisoner.postRecallReleaseDate)
 
-  private fun isApSledRelease(prisoner: PrisonerSearchPrisoner): Boolean = when {
+  private fun isBeingReleaseAtSled(prisoner: PrisonerSearchPrisoner): Boolean = when {
     prisoner.postRecallReleaseDate == null -> false
-
     prisoner.licenceExpiryDate == null -> false
 
-    prisoner.topupSupervisionExpiryDate != null && prisoner.topupSupervisionExpiryDate.isAfter(prisoner.licenceExpiryDate) -> false
```
andrewrlee, line 167: this would allow PSS licences if they were being released on SLED - now we don't allow any licences to be created on SLED

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/EligibilityService.kt
```diff
@@ -192,15 +190,7 @@ class EligibilityService(
...
 
-  private fun isPssTypeStillEligible(prisoner: PrisonerSearchPrisoner): Boolean {
-    if (prisoner.licenceExpiryDate == null &&
-      prisoner.topupSupervisionExpiryDate != null &&
-      isrPssProgressionService.isPssNowRepealed()
-    ) {
-      return false
-    }
-    return true
-  }
```
andrewrlee, line 203: if they're only PSS then exclude them.. this could be stricter - we could disallow people with no LED entirely but we don't currently so didn't want to add this in until after we talk about it

AledWEvans (approved):
LGTM - hopefully no other edge cases!

PR #1400, opened 2026-05-12
Cvsl 4036 cvl reinstate hdc variations

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/mapper/CurfewTimesMapper.kt
```diff
@@ -0,0 +1,28 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.mapper
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.CurfewTimes
+
+@Component
+class CurfewTimesMapper {
+
+  companion object {
+    fun copy(entity: CurfewTimes?): CurfewTimes? {
```
AledWEvans: If this is a spring component you should not put the copy in a companion Object as you are really using it as a static method and there is no need to have a @Componet! Components are for auto wiring and spring managment. Maybe just use a object like `object LicenceFactory` if you need static access!

AledWEvans (commented):
I think you need to tweak it here a bit

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/mapper/CurfewTimesMapper.kt
nishanthmahasamudram: updated.

AledWEvans (approved):
LGTM

PR #1401, opened 2026-05-13
CVSL-4103 Disable OSS Index analyzer in OWASP Dependency Check

nishanthmahasamudram:
**Fix: Disable OSS Index analyzer in OWASP check**
OWASP Dependency Check was failing due to missing/invalid OSS Index credentials, preventing the SARIF report from being generated.
Since the pipeline uses the NVD data feed, OSS Index is not required. This change disables it.


gradle.properties
```diff
@@ -1,3 +0,0 @@
-org.gradle.jvmargs=--illegal-access=permit
```
nishanthmahasamudram, line 1: unwanted OSS Index credentials

andrewrlee (approved):
LGTM

PR #1404, opened 2026-05-18
CVSL-4018: Add a field to specify if a COM is allocated or not to the…

justin-gardiner:
… response of get probation case.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/CaseService.kt
```diff
@@ -61,6 +62,18 @@ class CaseService(
     )
   }
 
+  fun getProbationAndAllocationInfo(prisonNumber: String): ProbationCaseModel {
+    val probationCase = getProbationCase(prisonNumber)
+    val offenderManager = deliusApiClient.getOffenderManager(probationCase.crn)
```
SebNegahbanMoJ: Does the result of `getOffenderManager` not contain the case information? I believe it has a `case` attribute which - at a glance - contains the information retrieved by `getProbationCase`

justin-gardiner: Ahh, I didn't look at it that way, I looked at the return from getProbationCase and couldn't see allocated so added the second call. I didn't check if the first call could be removed!. I'll check

gurps-singh (approved):

PR #1405, opened 2026-05-18
CVSL-4096 listen for supporting prison event

gurps-singh:
This PR is to add functionality to listen to an event where the supporting prison has changed for a restricted patient. This is triggered by a transfer from the Restricted Patients Service.


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/SupportingPrisonUpdatedHandler.kt
```diff
@@ -0,0 +1,109 @@
...
+
+      log.info("Processing supporting prison changed event received for nomis id: $nomisId")
+
+      val nomisRecord = prisonService.searchPrisonersByNomisIds(listOf(nomisId)).first()
+      if (!nomisRecord.isRestrictedPatient()) {
+        log.info("Nomis record is not a restricted patient, skipping update of supporting prison")
+        return
+      }
+
+      val relevantLicenceStatuses = LicenceStatus.PRE_RELEASE_STATUSES.toList() - NOT_STARTED
```
andrewrlee, line 54: don't need to worry about NOT_STARTED as we won't have any DBs in the DB that are NOT_STARTED

andrewrlee (approved):
LGTM

PR #1406, opened 2026-05-18
CVSL-4118 rename hdc licence id to licence version id

louisenorris:


src/main/resources/migration/common/V75__alter_hdcLicence_id_to_hdc_licence_version_id.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE hdc_migration_meta_data
+RENAME COLUMN hdcLicence_id TO hdc_licence_version_id;
```
louisenorris, line 2: **note:** checked online docs and when renaming a column any constraints that rely on the column name (there is a unique constraint on this column) are automatically updated 😅

AledWEvans (approved):
LGTM

PR #1412, opened 2026-05-21
CVSL-4039: Initial work to allow remand to be identified from the cou…

dave-virden:
…rt events.


run-local.sh
```diff
@@ -64,7 +64,7 @@ export HDC_ENABLED=true
 export STANDARD_RECALLS_ENABLED=true
 export RESTRICTED_PATIENTS_ENABLED=true
 export USE_CURRENT_HDC_STATUS=false
-export ISR_REPEAL_DATE="2026-03-25"
```
gurps-singh, line 67: **issue:** ‏Removed the ISR_REPEAL_DATE by accident and replaced it with the REMAND_ENABLED one.

dave-virden: Not sure what happened there but I've put it back

src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/remand/RemandIdentificationServiceTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+  @ParameterizedTest(name = "Returns true for outcome reason code {0}")
+  @ValueSource(
+    strings = [
+      "2507", "4001", "4004", "4012", "4016", "4505", "4506", "4531", "4532",
+      "4534", "4535", "4536", "4537", "4539", "4549", "4553", "4554", "4560",
+      "4561", "4563", "4564", "4565", "4570", "4571", "4588", "5601",
+    ],
+  )
+  fun `returns true when prisoner has a remand court event outcome`(outcomeReasonCode: String) {
```
andrewrlee, line 35: suggestion: could use an enumSource?

dave-virden: It could be, I've always done it this way to decouple the test from the enum so if you change the enum the test breaks unless you consciously fix it

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/remand/RemandIdentificationService.kt
```diff
@@ -0,0 +1,20 @@
...
+class RemandIdentificationService(
+  private val prisonApiClient: PrisonApiClient,
+) {
+
+  fun isRemandBooking(prisoner: PrisonerSearchPrisoner): Boolean = prisoner.bookingId?.let { bookingId ->
+    bookingsWithRemandOutcomes(listOf(bookingId.toLong())).isNotEmpty()
+  } ?: false
+
+  private fun bookingsWithRemandOutcomes(bookingIds: List<Long>): List<Long> {
+    val courtEventOutcomes = prisonApiClient.getCourtEventOutcomes(bookingIds, RemandCourtEvents.REMAND_CODES)
```
andrewrlee, line 17: we probably don't want to do another api call, if we're always going to call this at the same time as  we call it for IS-91 cases it'd be good to lump it in with that - might need a bit of refactoring to rename/rejig things

PR #1415, opened 2026-05-22
CVSL-4096 consume prisoner updated event for restricted patients

gurps-singh:
This PR is to use offender search events to recognise when a supporting prison ID has been updated by the restricted patients service. 

This is in response to initial work where we were listening for the nomis event but this led to a race condition where prisoner search was not updating with the correct supporting prison ID when we queried it. It also consumes the same event. 

After a chat with Andy, decided to go with the offender search event where a restricted patient record has changed. This means that the search record will be up to date to allow us to have the correct supporting Prison ID when we update licences.


dave-virden (approved):
LGTM

PR #1416, opened 2026-05-22
CVSL-4142 refactor getofcreatecom

louisenorris:
Fix to make sure a COM user is returned as this was previously failing some HDC licence migrations


AledWEvans (approved):
LGTM

PR #1418, opened 2026-05-22
CVSL-4096 remove log statement as no longer needed

gurps-singh:
Remove log statement after debugging testing issues for the supporting prison event consumption.


dave-virden (approved):
LGTM

PR #1426, opened 2026-06-01
An example of an abstract super class

AledWEvans:
Hi @nishanthmahasamudram here's an example of that abstract approach I would recommend for the address curfew changes... you should not really use interfaces as JPA does not really handle them well


nishanthmahasamudram:
This is not actual implementation.

PR #1432, opened 2026-06-02
CVSL-4194 HDC | Migration - work out address type and improve address…

AledWEvans:
follow on work from the hdc api side to take on the addressType


nishanthmahasamudram (approved):
LGTM

PR #1433, opened 2026-06-03
CVSL-3373: Add vehicle restrictions condition to policy V4.

justin-gardiner:


dave-virden (approved):
LGTM

src/test/resources/test_data/policy_conditions/policyV1.json
justin-gardiner, line 1: changed because of the new hideHintText = false

PR #1440, opened 2026-06-04
Code has several issues with it, had to refactor structure

AledWEvans:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -59,42 +62,53 @@ class HdcService(
...
+        ModelHdcCurfewAddress(
+          firstLine = it.addressLine1 ?: "",
+          secondLine = it.addressLine2,
+          townOrCity = it.townOrCity ?: "",
+          postcode = it.postcode ?: "",
+        )
+      }
     }
 
     val firstNightCurfewTimes = licence.firstNightCurfewTimes?.transformToModelFirstNightCurfewTimes()
```
AledWEvans, line 101: So this is strange why dose it ignore the firstNightCurfewTimes returned from hdc api from the hdcLicenceData object? I think this is a bug? @nishanthmahasamudram ?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/hdc/reponse/CurfewTimes.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.service.hdc.reponse
+
+import java.time.DayOfWeek
+import java.time.LocalTime
+
+data class CurfewTimes(
```
nishanthmahasamudram, line 6: Any reason why fromTime & untilTime is nullable?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -59,42 +62,53 @@ class HdcService(
...
+      }
     }
 
     val curfewAddress = if (licence.curfewAddress != null) {
       transformToModelHdcCurfewAddress(licence.curfewAddress!!)
     } else {
-      licenceData.curfewAddress
+      hdcLicenceData.curfewAddress?.let {
+        ModelHdcCurfewAddress(
+          firstLine = it.addressLine1 ?: "",
```
nishanthmahasamudram, line 93: Using "" might hide missing data and make it harder to distinguish between “not provided” and “intentionally empty”. Curious about the intended behaviour here.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/hdc/reponse/CurfewTimes.kt
AledWEvans: This is existing issues and can be solved separately

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
AledWEvans: After talking about this me and nish thing it's best to add these as empty as we do not what an exception to be thrown and prevent the user from seeing the state of the data from hdc, and possibly have the opportunity to correct. We can always change this later! Please read @andrewrlee

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/model/HdcCurfewAddress.kt
```diff
@@ -10,11 +10,9 @@ data class HdcCurfewAddress(
   val id: Long? = null,
 
   @field:Schema(description = "The first line of the curfew address", example = "1 Some Street")
-  @field:JsonProperty("addressLine1")
```
nishanthmahasamudram, line 13: looking good now.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
nishanthmahasamudram: No, actually we capture firstNightCurfewTimes from CVL UI.
So we can use it.

nishanthmahasamudram (approved):
LGTM

PR #1442, opened 2026-06-08
CVSL-4135 HDC | Migration - adding in lastPrisonId instead of prisonI…

AledWEvans:
…d for prison code - removed unnecessary fields from dto


louisenorris (approved):
LGTM

PR #1443, opened 2026-06-08
Cvsl 4076 add source to curfew address

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -94,6 +96,7 @@ class HdcService(
           secondLine = it.addressLine2,
           townOrCity = it.townOrCity ?: "",
           postcode = it.postcode ?: "",
+          source = AddressSource.MANUAL_MIGRATED,
```
nishanthmahasamudram, line 99: This will be always **MANUAL_MIGRATED** when we return address from HDC service.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/HdcService.kt
```diff
@@ -184,7 +187,7 @@ class HdcService(
   @Transactional
   fun addHdcCurfewAddress(licenceId: Long, request: AddHdcCurfewAddressRequest) {
     val licence = licenceRepository.findById(licenceId)
-      .orElseThrow { EntityNotFoundException("Hdc Licence $licenceId not found") } as HdcLicenceEntity
+      .orElseThrow { EntityNotFoundException("Hdc Licence $licenceId not found") } as HdcVariationLicence
```
nishanthmahasamudram, line 190: HDC curfew address will always be saved for HDC variations.

AledWEvans (approved):
LGTM

PR #1444, opened 2026-06-08
CVSL-4172 update mobile phone ban condition

gurps-singh:
This PR is to update the mobile phone ban condition for the 4th version of the licence conditions:

Screenshot 2026-06-08 at 15 18 45


dave-virden (approved):
LGTM

PR #1450, opened 2026-06-09
CVSL-4218 CVL | getting HDC Licence without curfew hours throws an ex…

AledWEvans:
A bug found during testing migration data, the bug is not the migration it self but the process afterwards, Fixing this will help us with our testing!


louisenorris (approved):
LGTM

PR #1469, opened 2026-06-15
CVSL-4201: Add policy version fo standard condition table and populat…

justin-gardiner:
…e when creating a licence.


andrewrlee (approved):
LGTM

PR #1473, opened 2026-06-16
CVSL-4228 increase memory size

gurps-singh:
This PR increases the maximum in memory size for the Delius client from 10MB to 12MB to stop alerts around a DataBufferLimitException and exceeding the maximum size.


dave-virden (approved):
LGTM

PR #1474, opened 2026-06-16
CVSL-4115 remove standard recall feature toggle

gurps-singh:
This PR is to remove the standard recall feature toggle from the backend.


justin-gardiner (approved):
LGTM

PR #1476, opened 2026-06-17
Cvsl 4184 vary do you want to use the standard hdc curfew hours page

nishanthmahasamudram:
Enable variation licence updates for HDC weekly curfew times API


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/HdcCurfewUpdatable.kt
```diff
@@ -0,0 +1,22 @@
+package uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity
+
+import uk.gov.justice.digital.hmpps.createandvaryalicenceapi.entity.Licence.Companion.SYSTEM_USER
+
+interface HdcCurfewUpdatable {
+
+  var weeklyCurfewTimes: MutableList<CurfewTimes>
+  var updatedByUsername: String?
+  var updatedBy: Staff?
```
andrewrlee: this could be added to HDC case?

andrewrlee: doesn't need to be a separate interface

andrewrlee (approved):
LGTM

PR #1485, opened 2026-06-18
CVSL-4186 Allowing HDC Variations to update curfew times

SebNegahbanMoJ:


SebNegahbanMoJ:
Turns out this has already been done, this is what I get for forgetting to pull in main first 😅

PR #1491, opened 2026-06-22
CVSL-4246 ensuring HDC child properites are copied

andrewrlee:


gurps-singh (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/entity/address/hdc/HdcCurfewAddress.kt
```diff
@@ -55,4 +55,22 @@ class HdcCurfewAddress(
   source = source,
   createdTimestamp = createdTimestamp,
   lastUpdatedTimestamp = lastUpdatedTimestamp,
-)
+) {
+
+  fun copy(licence: Licence) = HdcCurfewAddress(
+    id = id,
```
AledWEvans, line 61: Will this be copying it? because your passing the id, wont it just reuse it?
would have thought this needs to be set to Null!

PR #1492, opened 2026-06-22
CVSL-4082 add release label to vary caseload

gurps-singh:
This PR tweaks the LicenceComCase object to include homeDetentionCurfewEligibilityDate so that we can correctly discern the release labels on the vary caseload.


andrewrlee (approved):
LGTM

PR #1493, opened 2026-06-22
CVSL-4243 Support V4 electronic monitoring types in public api

andrewrlee:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies/Codes.kt
```diff
@@ -8,3 +11,52 @@ const val ELECTRONIC_TAG_COND_CODE_14E = "2F8A5418-C6E4-4F32-9E58-64B23550E504"
 const val MULTIPLE_UPLOAD_COND_CODE = "005d70e4-a247-4f82-b8b3-6d294a0f5051"
 const val EXCLUSION_ZONE_COND_CODE = "0f9a20f4-35c7-4c77-8af8-f200f153fa11"
 const val EVENT_EXCLUSION_COND_CODE = "99195049-f355-46fb-b7d8-aef87a1b19c5"
+
+enum class Version {
+  V1,
+  V2,
+}
+
+enum class ElectronicMonitoringType(val text: Map<Version, String?>) {
```
AledWEvans, line 20: String? can this be null?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies/Codes.kt
```diff
@@ -8,3 +11,52 @@ const val ELECTRONIC_TAG_COND_CODE_14E = "2F8A5418-C6E4-4F32-9E58-64B23550E504"
...
+    V2 to "that you do not leave areas you must stay in (restriction zones)",
+  ),
+  ;
+
+  constructor(vararg values: Pair<Version, String?>) : this(values.toMap())
+
+  companion object {
+    fun find(value: String): ElectronicMonitoringType? = ElectronicMonitoringType.entries.find { it.text.values.contains(value) }
+  }
+}
```
AledWEvans, line 62: TBO I find this confusing  as Policy V1 and V2 have a different aproach to V3 and V4 would be nice if they used the same ElectronicMonitoringType approach, but I think I also understand why you have done it this way. I like consistency!

AledWEvans (commented):
Some minor comments your call!

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies/Codes.kt
andrewrlee: yes, it can

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/policies/Codes.kt
andrewrlee: I didn't want to use policy version as we'd have to add a new entry for every policy version which would be a bit confusing if the condition doesn't change..

AledWEvans (approved):
LGTM

PR #1498, opened 2026-06-24
CVSL-4264 do not get CVL record for variations

andrewrlee:


run-local.sh
```diff
@@ -47,6 +47,8 @@ export DB_SERVER=localhost
 export DB_NAME=create-and-vary-a-licence-db
 export DB_USER=cvl
 export DB_PASS=cvl
+export PROGRESSION_MODEL_POLICY_START_DATE=2026-01-14
+export HDC_ENABLED=true
```
andrewrlee, line 51: Added these as most people are working with HDC now

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -440,40 +440,36 @@ class LicenceService(
     val submitter = staffRepository.findByUsernameIgnoreCase(username)
       ?: throw ValidationException("Staff with username $username not found")
 
-    val nomisRecord = prisonerSearchApiClient.searchPrisonersByBookingIds(listOf(licenceEntity.bookingId!!)).first()
```
andrewrlee, line 443: this call fails intermittently for variations but isn't actually used by them

AledWEvans (commented):
LGTM, nice fix and refactor also, assume exiting tests cover this

AledWEvans (approved):

PR #1511, opened 2026-07-02
CVSL-4277 change em report to only bring back cases with LSD today or…

gurps-singh:
… in the future

This PR is to amend the EM report to match the pre release report showing those cases where the LSD is today or in the future only.


justin-gardiner (approved):
LGTM

PR #1512, opened 2026-07-02
CVSL-4301 HDC cases appearing in the vary caseload

AledWEvans:
you could argue this is not needed as we should not have hdc licences on system but what if we roll back!


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/caseload/com/ComVaryCaseloadService.kt
```diff
@@ -74,9 +76,8 @@ class ComVaryCaseloadService(
       }
       when {
         licence == null -> null
-
+        !hdcEnabled && licence.kind.isHdc() -> null
```
nishanthmahasamudram, line 79: Can we add integration test to this?

nishanthmahasamudram (approved):
LGTM

PR #1516, opened 2026-07-02
CVSL-4307 Prog HDC | CVL UI | Standard conditions are missing from th…

AledWEvans:
…e HDC licence template at all times


nishanthmahasamudram (approved):
LGTM

PR #1520, opened 2026-07-06
CVSL-4311 Prog HDC | Lifecycle data not migrated across following var…

AledWEvans:
corrisponding changes for the api change in hdc


andrewrlee (approved):
LGTM

PR #1522, opened 2026-07-06
CVSL-4288: Rename appointment person type to appointment type and add…

justin-gardiner:
… no appointment needed option.


andrewrlee (approved):
LGTM

PR #1527, opened 2026-07-08
CVSL-4341 Removing deprecated appointmentContact field as it's no lon…

andrewrlee:
…ger used


AledWEvans (approved):
LGTM assuming your leaving the column in the DB!?

PR #1533, opened 2026-07-14
CVSL-4156 remove old creation endpoint

gurps-singh:
This PR is to remove the old licence creation endpoint which is no longer in use.


justin-gardiner (approved):
LGTM

PR #1534, opened 2026-07-14
Cvsl 2048 prisoner merged event

justin-gardiner:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/resource/privateApi/SupportController.kt
```diff
@@ -205,4 +210,42 @@ class SupportController(
   fun syncComAllocation(
     @PathVariable crn: String,
   ) = comAllocatedHandler.syncComAllocation(crn)
+
+  @PutMapping(value = ["/merge"], produces = [MediaType.APPLICATION_JSON_VALUE])
```
justin-gardiner, line 214: So we have a way of testing without wiping out prisoners in nomis.

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/domainEvents/PrisonerMergedHandler.kt
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  fun mergeOffenders(oldNomisId: String, newNomisId: String, newBookingId: Long) {
+    val licencesToUpdate =
+      licenceRepository.findAllByNomsIdAndStatusCodeIn(oldNomisId, IN_FLIGHT_LICENCES)
+    if (!licencesToUpdate.isEmpty()) {
+      val (oldBookingLicences, newBookingLicences) = licencesToUpdate.partition { it.bookingId != newBookingId }
+      deactivateLicencesOnOldBooking(oldBookingLicences)
+      updateOffenderDetails(newBookingLicences, newNomisId)
+    }
```
justin-gardiner, line 53: After discovering the was a domain event and not prison event, I discussed with Andy and we decided to not update supporting prison here as it's only available in prisoner search and would lead to a race condition. It should be picked up by the prisoner updated event.

gurps-singh (approved):

PR #1539, opened 2026-07-16
Prisoner name is converted to title case.

AledWEvans:
Migrate update code to create title case to prisoner names


justin-gardiner (approved):
LGTM

PR #1541, opened 2026-07-16
CVSL-4362 change tpl field as well

gurps-singh:
This PR contains some missing config I forgot to add for the updating of the condition text for curfew.


dave-virden (approved):
LGTM

PR #1558, opened 2026-07-23
CVSL-2048: Update prisoner merged query to get all existing licences …

justin-gardiner:
…instead of just inflight.


andrewrlee (approved):
LGTM

PR #1560, opened 2026-07-23
CVSL-4344 Adding job to deactivate progression licences on old policy…

SebNegahbanMoJ:
… versions


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/DeactivateProgressionLicencesService.kt
```diff
@@ -0,0 +1,107 @@
...
+          surname = "SYSTEM",
+          eventDescription = "Licence automatically deactivated as it is a progression licence on an older policy version for ${licence.forename} ${licence.surname}",
+        ),
+      )
+      domainEventsService.recordDomainEvent(licence, LicenceStatus.INACTIVE)
+
+      // query relies on Licence Start Date being past given date, so nulls should not be returned
+      if (licence.licenceStartDate!!.isOnOrBefore(notificationWindowEndDate)) {
+        if (licence.responsibleCom == null) {
+          log.info("Unable to notify COM of progression licence deactivation as licence has no responsible com")
```
andrewrlee: need to add some context to this so we know which licence is affected?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/DeactivateProgressionLicencesService.kt
```diff
@@ -0,0 +1,107 @@
...
+          emailAddress = licence.responsibleCom?.email,
+          crn = licence.crn!!,
+          comFirstName = licence.responsibleCom!!.firstName!!,
+          comLastName = licence.responsibleCom!!.lastName!!,
+          pipFirstName = licence.forename!!,
+          pipLastName = licence.surname!!,
+        )
+      }
+    }
+    telemetryService.recordDeactivateProgressionLicencesJobEvent(licences.size)
```
andrewrlee: style thing - would move this up to the deactivateLicences function as it's at a similar level of abstraction

SebNegahbanMoJ: We may have to move this in other places too as I've copied a pattern we use elsewhere, but I've got no issues with doing it here

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
```diff
@@ -210,4 +210,15 @@ interface LicenceRepository :
...
+  @Query(
+    """
+    SELECT l
+    FROM Licence l
+    WHERE l.version in ("1.0", "2.0", "2.1", "3.0")
+    AND l.statusCode in (uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.TIMED_OUT, uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.IN_PROGRESS, uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.SUBMITTED, uk.gov.justice.digital.hmpps.createandvaryalicenceapi.util.LicenceStatus.APPROVED)
+    AND l.licenceStartDate >= :policyV4GoLiveDate
+  """,
+  )
+  fun getLicencesForProgressionDeactivation(policyV4GoLiveDate: LocalDate): List<Licence>
```
andrewrlee, line 223: we have a slightly odd progression repository that this could go on.. 
there's a bit of inconsistency, either that could be merged to here or this should go there?

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/DeactivateProgressionLicencesService.kt
```diff
@@ -0,0 +1,107 @@
...
+@Service
+class DeactivateProgressionLicencesService(
+  private val licenceRepository: LicenceRepository,
+  private val auditEventRepository: AuditEventRepository,
+  private val licenceEventRepository: LicenceEventRepository,
+  private val domainEventsService: DomainEventsService,
+  private val telemetryService: TelemetryService,
+  private val notifyService: NotifyService,
+
+  @param:Value("\${progression.model.policy-start-date}")
```
andrewrlee: it'd be worth updating this to handle nulls: https://github.com/ministryofjustice/create-and-vary-a-licence-api/pull/1574

src/main/resources/application.yml
```diff
@@ -179,6 +181,11 @@ feature:
     hdcCreation:
       enabled: ${HDC_CREATION_ENABLED:#{false}}
 
+progression:
+  model:
+    policy-start-date: ${PROGRESSION_MODEL_POLICY_START_DATE}
```
andrewrlee: ah.. if we're including this - we should handle nulls here, similar to the `${HDC_CREATION_ENABLED:#{false}}` example above

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
SebNegahbanMoJ: I'd argue for merging it in here, as the queries are being performed on the licences table and returning licence objects. The only difference is the context in which the results are used, which feels somewhat like not the repository's concern

andrewrlee: I agree

helm_deploy/create-and-vary-a-licence-api/values.yaml
```diff
@@ -124,3 +124,15 @@ jobs:
     path: /notify-probation-of-unapproved-licences
     cron: "0 2 * * *" # 2am every day
     allowRetry: false
+
+  - name: isr-licence-progression-job
+    enabled: true
+    path: /isr-licence-progression-job
+    cron: "30 0 11 5 *" # 00:30 UTC on 11/05/2026
+    allowRetry: true
```
andrewrlee: this needs to be removed now

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/repository/LicenceRepository.kt
andrewrlee: I'll do this as a separate change

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/DeactivateProgressionLicencesService.kt
```diff
@@ -0,0 +1,113 @@
...
+  @param:DateTimeFormat(pattern = "yyyy-MM-dd")
+  private val policyV4StartDate: LocalDate?,
+
+  @param:Value("\${progression.model.notification-window-end-date}")
```
andrewrlee: both these need to handle null syntax: 
e.g: `@param:Value("\${progression.model.notification-window-end-date:#{null}}")`

> Both fields will actually crash on startup if the property is absent, because Spring can't inject a missing `@Value` into a non-nullable-at-injection-time field — the LocalDate? nullability is Kotlin-level, but Spring will throw before that matters.

helm_deploy/create-and-vary-a-licence-api/values.yaml
SebNegahbanMoJ: I feel like I'm going nuts, I'm certain I did

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/DeactivateProgressionLicencesService.kt
SebNegahbanMoJ: Gotcha, thank you

src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/jobs/DeactivateProgressionLicencesService.kt
```diff
@@ -0,0 +1,113 @@
...
+
+        notifyService.sendLicenceDeactivatedForProgressionEmail(
+          emailAddress = licence.responsibleCom?.email,
+          crn = licence.crn!!,
+          comFirstName = licence.responsibleCom!!.firstName!!,
+          comLastName = licence.responsibleCom!!.lastName!!,
+          pipFirstName = licence.forename!!,
+          pipLastName = licence.surname!!,
+        )
+      }
```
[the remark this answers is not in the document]
SebNegahbanMoJ: Contextually, I'm reasonably confident that if a responsible COM exists then these fields should be defined. The same for the licence

andrewrlee (approved):
LGTM

PR #1562, opened 2026-07-24
Cvsl 4380 show before and after hdc curfew hours audit event

gurps-singh:
This PR is to change the layout of the audit event when curfew hours change so that it is easier to understand. 

Screenshot 2026-07-24 at 10 32 15


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/AuditService.kt
```diff
@@ -446,6 +449,17 @@ class AuditService(
     .findAllByEventTimeBetweenOrderByEventTimeDesc(auditRequest.startTime, auditRequest.endTime)
     .transformToModelAuditEvents()
 
+  private fun setAuditEventChangesOrder(event: ModelAuditEvent): ModelAuditEvent {
+    val changes = event.changes?.takeIf { it["type"] == UPDATED_HDC_CURFEW_TIMES_TYPE } ?: return event
+
+    val orderedChanges = objectMapper.convertValue(
```
gurps-singh, line 455: **note:** ‏A nifty Copilot suggestion which firstly coverts the map into the DTO so Jackson knows the structure and order and then converting this DTO back to a map for the existing changes field.

dave-virden (approved):
LGTM

PR #1567, opened 2026-07-27
CVSL-4378 CVL | Add audit event to log when HDC migration occurs - it really should be a SYSTEM_EVENT

AledWEvans:
should be a SYSTEM_EVENT


andrewrlee (approved):
LGTM

PR #1573, opened 2026-07-28
CVSL-4388 Enhance-Audit-Logging-for-Variation-Discard-Actions

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/LicenceService.kt
```diff
@@ -870,13 +870,15 @@ class LicenceService(
     val username = SecurityContextHolder.getContext().authentication?.name!!
     val discardedBy = this.staffRepository.findByUsernameIgnoreCase(username)
 
+    val auditLicenceId = if (licenceEntity is Variation) licenceEntity.variationOfId else licenceId
```
andrewrlee, line 873: we can probably rework this function to be specifically for variations - but out of scope of this ticket

andrewrlee (approved):
LGTM

PR #1577, opened 2026-07-29
 CVSL-4357 inactivate-V4-for-null-policy-start-date

nishanthmahasamudram:


src/main/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/UpdateSentenceDateService.kt
```diff
@@ -273,31 +274,22 @@ class UpdateSentenceDateService(
...
-  private fun crossesPolicyVersionCutoff(
-    oldLsd: LocalDate,
-    newLsd: LocalDate,
-    cutoffDate: LocalDate,
-  ): Boolean = !oldLsd.isBefore(cutoffDate) && newLsd.isBefore(cutoffDate)
+  private val Licence.isEligibleForPolicyVersionCheck: Boolean
+    get() = version == V4_0.version && statusCode in LicenceStatus.PRE_RELEASE_STATUSES
+
+  private val DateChange.newValidDate: LocalDate?
+    get() = takeIf { it.changed }?.newDate
```
andrewrlee: I think similarly here we probably don't need to worry if the date has changed?

andrewrlee: if the new date is prior to policy cut off we can assume it shouldn't be v4 so we can inactivate it.

nishanthmahasamudram: Yup will remove this check as well.

build.gradle.kts
```diff
@@ -147,6 +147,7 @@ tasks {
     filter {
       excludeTestsMatching("*.integration.*")
     }
+    jvmArgs("-Xmx2g")
```
nishanthmahasamudram: Maximising heap memory for Unit test, Int test and SAR snapshot generation.

andrewrlee (approved):
LGTM

PR #1579, opened 2026-07-29
CVLS-4343 remove deprecated comname and comstaffcode

louisenorris:


AledWEvans (approved):
LGTM, assume front end  does not reference any of these

PR #1585, opened 2026-08-06
CVSL-4407 - change exceptions when missing com to not retry

AledWEvans:
Changed exception type to not retry when missing com, most of the com issues are an issue with preprod,.. and if they occur in live we dont want them to push the message on DLQ


dave-virden (approved):
LGTM

